PR #5, opened 2021-07-31
Windows: restore SwiftInstaller Custom Action

compnerd:
This restores the SwiftInstaller custom action, relicensed under the
Apache 2.0.  The third party dependency on the Microsoft code is now
managed via nuget, and will be fetched as appropriate when building.


compnerd:
CC: @shahmishal

PR #14, opened 2021-08-05
Windows: add ICU packaging rules

compnerd:
We currently split up the ICU packaging as it is not part of the
toolchain but a required dependency and the version must match.


compnerd:
CC: @shahmishal

shahmishal (approved):

PR #30, opened 2021-10-02
build ninja from source

tomerd:
motivation: more tightly control the version of ninja used

changes:
* add download instricutions for ninja source
* remove present patch since we are building ninja from source
* cleanup tarball names


shahmishal:
Similar work I was doing locally but also dropping all patch files - https://github.com/apple/swift-installer-scripts/pull/31

shahmishal (approved):

PR #23, opened 2021-09-17
debian: add initial control files for debian packaging

compnerd:
The packaging on debian mostly requires the control files and a
pre-configured layout to package as a root.  This adds the control files
as a starting point.

PR #33, opened 2021-10-06
externalize metadata and make it version based

tomerd:
motivation: share package metadata for swift versions and platforms

changes:
* introduce new directory for share metadata
* create metadata file for rpms, swift 5.5.0
* update centos 8 spec and script to use the shared metadata


tomerd:
@shahmishal this builds on top of #32, see https://github.com/apple/swift-installer-scripts/pull/33/commits/31db6d335e420cf7d83f542fccfdb749f1f0b334 for the actual changes, once we merge #32 I will rebase on top

tomerd:
@shahmishal @tachoknight @futurejones wdyt about this _technical_ approach to achieve some of the ideas mentioned inhttps://github.com/apple/swift-installer-scripts/pull/23#issuecomment-938275941? we still need to figure out the actual values, so focus on the technical aspect of how this is wired up

tachoknight:
> I really think that this project should be separated into the different package manager families of distributions. 

This seems like just an additional level of indirection; if I want to build Swift for Ubuntu or Fedora, why not just go to the appropriate directory directly, without having to be know about the package type?

futurejones:
@tachoknight, what I am discussing is not about building swift. Its about packaging the swift toolchain for distribution using the native packaging system.
Knowing the package type and what the different package managers require is the key point and the most relevant information. I wouldn't have thought of this as an 'additional level of indirection'.
If you want to build the swift toolchain then this is already covered at https://github.com/apple/swift.

tomerd:
@futurejones i like the proposed directory structure and totally agree with your assessment 

note this PR does not attempt to canonicalize the metadata across different package managers - the proposed shared metadata file is qualified to be for RPM only, so it can be shared across centos, amazon linux, etc

tomerd:
> what I am discussing is not about building swift. Its about packaging the swift toolchain for distribution using the native packaging system.
Knowing the package type and what the different package managers require is the key point and the most relevant information. I wouldn't have thought of this as an 'additional level of indirection'.

i suspect @tachoknight point was the explicit directory representing the type of package (eg RPM) is redundant as it can be implicitly derived from the linux distro 

personally i can go either way

tomerd:
@futurejones @tachoknight @shahmishal updated to reflect the directory structure suggested by @futurejones which I think makes this nice.

should we move forward with this?

tachoknight:
Admittedly I'm not sure how it makes it better than just going to the appropriate distro directory, but if that's how folks like it, then ¯\_(ツ)_/¯.

tomerd:
@shahmishal please merge when you are ready

soloturn:
what i'd really would have loved this cool metadata idea would solve, that the dependencies can be easier extracted. e.g.
- a list of software and version which the build depends on, e.g. ninja, ICU, i.e. non-apple
- a list of urls to build swift, e.g. llvm-project
- a list of urls to be built when swift is built, swift source code, e.g. swift-crypto
- the list of URLs in representation git-url, maybe even a git repo with submodules so one can do a shallow clone with depth 1.
- the list of URLs in represenation binary download

because currently i do the following check update_checkout/update-checkout-config.json, and need to check if it is tag or branch as arch wants the direct https url to feed into "git clone".

tomerd:
> @shahmishal @tomerd for the arch linux and family build what @tachoknight suggests seems slimmer. the package scripts are so small that a directory structure imo over-complicates it.

it makes it easier to share metadata in a format that is agnostic to the package manager. the original PR had it in a separate place, but this makes it easy to reason about. in any case, this is something we can change later, so I suggest we continue with this for the sake of making progress 

> what i'd really would have loved this cool metadata idea would solve, that the dependencies can be easier extracted. e.g.
> a list of software and version which the build depends on, e.g. ninja, ICU, i.e. non-apple
> a list of urls to build swift, e.g. llvm-project
> a list of urls to be built when swift is built, swift source code, e.g. swift-crypto
> the list of URLs in representation git-url, maybe even a git repo with submodules so one can do a shallow clone with depth 1.
> the list of URLs in represenation binary download
> because currently i do the following check update_checkout/update-checkout-config.json, and need to check if it is tag or > branch as arch wants the direct https url to feed into "git clone".

this is the general direction I would like to go to, but this PR starts with the minimal and easy to extract and share metadata. I can see follow on PRs to extract these as well, especially since some of this may be version specific

soloturn:
> > @shahmishal @tomerd for the arch linux and family build what @tachoknight suggests seems slimmer. the package scripts are so small that a directory structure imo over-complicates it.
> 
> it makes it easier to share metadata in a format that is agnostic to the package manager. the original PR had it in a separate place, but this makes it easy to reason about. in any case, this is something we can change later, so I suggest we continue with this for the sake of making progress

agreed, metadata should be as close as possible to the usage. but the structure seems really complex, let us digest this path: ```platforms/Linux/RPM/shared/metadata/5.5.0/metadata.inc```.  why it cannot be simply: ```rpm/metadata.inc```, and ```rpm/fedora-8/package.spec```, ```msi/whatebber``` ?

lets then look at why rpm should be there. you have the problem of operating systems where one could have multiple formats, e.g. for windows you have msi, chocolatey, winget, and zip. for fedora you have rpm, and tar ball. arch linux base, use shell scripting, just like tar ball.

another decision point is the version number. 5.5.0 is in the path while the whole structure is stored in a source code repo, where versions are usually tags, not in the path. of course there may be good reasons for doing it. a new version of swift-formater comes, and you want to change it for 5.5.x and 5.4.x in one go instead of checking out 20 branches and re-tag. but IFF you decide that this repo should not follow the same versioning schemes as other swift repos, why not making it so that you can easy drop prehistoric stuff easily later on, and put the version on top: ```5.5.0/rpm/metadata.inc```.

another aspect is the change rate. ideally this metadata contains volatile data like version numbers which is a pain. but - here i agree with you that it can be fixed any time.

in a summary, a simple directory structure would be cool. extracting later on is easy in case somebody finds it too much work to maintain. something flexible like: 
```
swift-installer-scripts
  ├ fedora-8
  │   ├ package.spec
  │   └ ...
  ├ shared
  │   ├ rpm
  │   │   ├ metadata.inc
  │   │   └ ...
  │   └ ...
  ├ windows-11
  │   ├ winget
  │   │   ├ swiftlang.json
  │   │   └ ...
  │   ├ msi
  │   └ ...
  └ ...
```
that fedora uses the rpm shared is convention. if you really like it, put ```5.5.0``` on top.

PR #34, opened 2021-10-08
Changes to make Fedora more similar to CentOS scripts

tachoknight:
This PR makes the Fedora scripts more similar to the CentOS one. I moved patch files to a separate directory, incorporated Ninja as part of the build and removed it from the build requirements, and updated the README file to incorporate the changes.


tomerd:
seems good to me, consider maybe jumping straight to the docker-compose setup?

PR #36, opened 2021-10-08
improve docker setup

shahmishal:
Reverts apple/swift-installer-scripts#35


tomerd:
ci would need to be changed to do the following:

```
cd <path>
# update the base image if needed
docker-compose build --pull
# build the package in <path>/.output
docker-compose run build
```

PR #37, opened 2021-10-12
Update the README with Linux package / installer info (RPM/DEB)

shahmishal:


shahmishal:
cc: @tomerd @tachoknight @futurejones @compnerd 

I am planning on creating the bugs.swift.org tickets to track the work after we review the README.

tomerd (approved):

README.md
```diff
@@ -25,6 +25,71 @@ swift-installer-scripts
...
+- [ ] SR-### Setup CI to build the deb package
+- [ ] SR-### Code sign package with swift.org certificate
+- [ ] SR-### Host the deb package on swift.org
+- [ ] SR-### Host the deb repository on swift.org
+- [ ] SR-### Verify the deb package and repository
+- [ ] SR-### Update swift.org download page
+- [ ] SR-### Work with official repositories to accept package specs
+
+*For each platform, we will start with Ubuntu 20.04
+
```
tomerd, line 92: should we add task(s)
to update swift.org install page for the available packages?

README.md
```diff
@@ -25,6 +25,71 @@ swift-installer-scripts
...
+  * Support all officially supported Linux platforms
+  * Code signed by swift.org certificate
+  * Repository hosted on swift.org
+* Step 3. Offer the native packages / installer through official repositories
+for the various platforms
+  * Work with official repositories to accept package specs
+  * Deprecate swift.org packages / installer repository
+* Step 4. Deprecate swift.org Linux tarballs
+
+### Package Info
```
tomerd, line 47: @futurejones @tachoknight @compnerd please review and see if you have any feedback on the suggested metadata here (this is a draft)

README.md
```diff
@@ -25,6 +25,71 @@ swift-installer-scripts
            └ ...
 ~~~
 
+## Linux Packages (RPM/Deb)
```
tomerd, line 28: few other things we need to agree on and document are:
1. package for development (full toolchain) vs. package for runtime (supporting libraries for swift applications)
2. location  of the swift binaries. eg do we go in /opt or /user
3. how would we version the packages, especially in light of lack of stable abi on linux? is minor version and being careful with patch versions enough, or should we issue a package per patch version?

tomerd: cc @futurejones @tachoknight @compnerd @lukasa @weissi @millenomi

shahmishal: > package for development (full toolchain) vs. package for runtime (supporting libraries for swift applications) 
We should have two packages: 
* `swiftlang` - contains both the compiler and runtime 
* `swiftlang-runtime` - contains only runtime

However, we should focus on `swiftlang` first and work on `swiftlang-runtime` after we have setup `swiftlang`. 

>how would we version the packages, especially in light of lack of stable abi on linux? is minor version and being careful with patch versions enough, or should we issue a package per patch version?

We should use the full version including the patch version until we are ABI stable on Linux platform.

futurejones: > We should use the full version including the patch version until we are ABI stable on Linux platform.

Even after ABI stability, patch versions will still be needed otherwise `apt update && upgrade` will not pull down and install the latest available version. e.g. updating from `5.4.2` to `5.4.3`.

shahmishal: > > We should use the full version including the patch version until we are ABI stable on Linux platform.
> 
> Even after ABI stability, patch versions will still be needed otherwise `apt update && upgrade` will not pull down and install the latest available version. e.g. updating from `5.4.2` to `5.4.3`.

Agree, the version has to fully match with swift release.

Tom is asking regarding the version number in the path/binary name. In other words, how do we want to handle multiple Swift version on the same system? 

For example: 
/usr/bin/swift-5.4.3 
/usr/lib/swift-5.4.3/

`update-alternatives --install /usr/bin/swift swift /usr/bin/swift-5.4.3 100`

tachoknight: > > 1. location of the swift binaries. eg do we go in /opt or /usr
> 
> With deb packages I have found using the `/usr` directory works best for addressing conflicts with existing native installs such as `clang`, `llvm`, `lldb` , `libicu` etc. It also makes it quick and simple to create the deb packages from the existing swift.tar.gz files as they are already using `/usr`.

For the "official" RPM for Fedora I put the entire toolchain in `/usr/libexec/swift` and symlink `swift`, `swiftc`, and `sourcekit-lsp` into `/usr/bin`; this keeps the Swift `clang` and `lldb` out of the path and `libexec` is generally regarded as a place to store programs meant to be called by other programs, so I figured that was the best place for the entire Swift toolchain, while preserving the `/usr /lib etc` structure that the toolchain expects.

tachoknight: > few other things we need to agree on and document are:
> 
> 2. location  of the swift binaries. eg do we go in /opt or /usr

FYI, for Fedora no package may be installed in `/opt` from an "official" RPM as matter of policy.

> 3. how would we version the packages, especially in light of lack of stable abi on linux? is minor version and being careful with patch versions enough, or should we issue a package per patch version?

I take the approach of packaging each `-RELEASE` version, regardless of major or minor versions.

shahmishal: Summary: 

- Multiple packages: We should handle this after we get all of the `swiftlang` RPM/Debs configured. `swiftlang` will contain both compiler and the runtime. 
- Location: We need to decide between` /usr/libexec/ ` vs ` /usr/`, because /opt is not valid option for Fedora. 
- Multiple versions on the system: We should revisit this in the future. 

@tomerd @tachoknight @futurejones @Lukasa I would like to merge this PR, however we need to decide on the install location.

stevapple: > * Location: We need to decide between`/usr/libexec/` vs ` /usr/`, because /opt is not valid option for Fedora

I don’t know about RedHat-based distros very much, but on Debian-based distros `/opt` is far more suitable than the other two. And according to @tachoknight ’s explanation I would suggest `/usr/libexec` for Fedora/CentOS.

Could we have these two different paths for different distros?

futurejones: @tomerd @shahmishal 
> * Location: We need to decide between/usr/libexec/ vs /usr/, because /opt is not valid option for Fedora

There is a big missing part in this discussion about the location the installation. The first question should be why?
The correct location for the swift install is `/usr`.
This should not be up for debate.
The reason why we are discussing it is because of conflicts caused by the inclusion of other packages in the swift toolchain.
So the entire reason for moving the install location from `/usr` is a work around to fix problems caused by the current state of the swift toolchain.
The main problems are caused by the inclusion of the entire llvm-project, icu and an erroneous python-six file.
The current conflicting packages are - 

- clang
- llvm
- lld
- lldb
- libicu
- python-six

There may be more which haven't been found yet.

Without these conflicts there would be no reason to install swift anywhere other than `/usr`

The solution should not be to just hide everything in a separate directory and forget about it. 
Ideally we should be separating out all the different packages and using the native versions where possible. 
Where not, a custom version can be built and included in the `swift repository`.
The swift package deb/rpm would then include this custom version as a dependency in the spec/control file. 

In the short term there is a `replaces` feature which allows the included packages in the swift install to replace any existing conflicting packages.
This is the best short term solution and will it allow a staged separation of packages over time without changing the user installation experience.

In regard to the current location choices, while both locations will technically work, I feel that neither should be used.
`/opt` - this is not what this location is designed for. 
`/libexec` - should only be used for applications which are not being directly accessed by the user.

futurejones: @tachoknight was there a particular reason why you couldn't replace the installed version of `lldb` with the swift version?

shahmishal: Propose we setup a meeting to discuss this, and figure out the next steps. @futurejones @tachoknight @compnerd @Lukasa @weissi @millenomi @stevapple @tomerd. 

We have Engineering Operations Sync meeting Oct 15th 2021 at 10:30am PST, If you are interested in joining please let me know I will send you the Webex info.

compnerd: I believe that `/opt` is the right place for the swift packages to install to for the forseeable future.  Until the llvm-project conflict is resolved, we have a conflict with a large number of packages.  By placing the package into /opt, we can have co-installation of the packages and the user can simply alter the path to get either the swift packages or the system packages for clang and the rest of the binary tools from llvm.  I don't think that the llvm-project fork is something that will be resolved in a month or so, so, for that duration we should prefer /opt as if Swift was a binary package that was shipped by a vendor.  Once the llvm-project fork issue has been resolved, we could easily work to split up the toolchain into separate clang/llvm/lld/lldb packages and provide newer versions of the package as dependencies for Swift.  At that time, we would have a natural motivation and point for transitioning from `/opt` to `/usr`.

README.md
compnerd: I think that we should consider the renaming policy for Debian: https://wiki.debian.org/RenamingPackages.  It may still be possible to use swift as the package name.  While it is common for some languages (e.g. go) to use the lang suffix, it generally goes by golang and rather than go.  Rust does something similar: rust is the package name and provides an uber package (rust-all) for installing all the pieces together.

README.md
shahmishal: Summary from the meeting we had today, we will need to have a follow up meeting. 

Attendees: @tomerd @tachoknight @compnerd @shahmishal 

Location:
Option 1: Diverge the install location between platform to best fit the platform requirements. 
    - symlink the toolchain into /usr/ to avoid conflicting with llvm.org binaries.
Option 2: Install in /usr and rename llvm-project binaries (example: swift-lldb/lldb-swift ...) 
Option 3: Install in /opt however this will not for work fedora, and we will still have issue with llvm-project binaries. 

We did not get to the discussion about multiple version or multiple packages (runtime vs full package). 

@tomerd @tachoknight @compnerd feel free to add anything I missed.

stevapple: I would prefer Option 1 — Until we finally manage to separate out all the dependencies.

---

For multiple versions and packages,

```
swiftlang
 -> swiftlang5
     -> swiftlang5.6
         -> swiftlang5.6-runtime
         -> swiftlang5.6-...
         -> ...
```

may be a possible naming convention. This will allow user to keep any `x.y` releases of Swift, or to keep up with the latest toolchain.

tachoknight:
I'd suggest the forums; it's pretty easy to get lost in GitHub's PR interface, and the discussion may span multiple PRs.

PR #38, opened 2021-10-14
add amazon linux 2 spec

tomerd:
motivation: support amazon linux 2

changes: add RPM spec file and supporting machinary for amazon linux 2


tomerd:
@shahmishal this is ready

PR #50, opened 2021-11-02
Add Deb package convention

stevapple:
A few points to mention:
- Apt has native support for `release+branch`, so the repos are better arranged in `OS` instead of `OS_VERSION`. This is different from `yum`.
- The conventional path for the package is too awkward for manual downloads.
  - A usual solution is to provide the package through symbolic links (3xx redirection?).
- Apt has a different naming convention from Swift.
  - For OS version, code names like `focal` are **preferred** over `20.04` and `2004`. 
  - For Arch, `amd64` and `arm64` are **used** instead of `x86_64` and `aarch64`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,13 +82,17 @@ Repository configuration: `/repo/<OS>/dists/<OS_VERSION>/swiftlang.list`
 ```
 /repo/centos/releases/8/x86_64/swiftlang-5.5.0-1.el8.x86_64.rpm
 /repo/centos/releases/8/aarch64/swiftlang-5.5.0-1.el8.aarch64.rpm
+/repo/ubuntu/pool/main/s/swiftlang/swiftlang-5.5.0-1-focal_amd64.deb
+/repo/ubuntu/pool/main/s/swiftlang/swiftlang-5.5.0-1-focal_arm64.deb
 ```
 
 * **Package URL:**  
 https://download.swift.org/repo/centos/releases/8/aarch64/swiftlang-5.5.0-1.el8.aarch64.rpm
+https://download.swift.org/repo/ubuntu/pool/main/s/swiftlang/swiftlang-5.5.0-1-focal_arm64.deb
 
 * **Repository configuration file URL:**  
 https://download.swift.org/repo/centos/releases/8/swiftlang.repo
+https://download.swift.org/repo/ubuntu/dists/focal/swiftlang.list
 
 ## Tasks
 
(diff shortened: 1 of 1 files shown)


shahmishal:
cc: @tomerd @compnerd @futurejones @tachoknight

compnerd:
Yeah, I agree with @stevapple here.  The layout being documented here seems far more reasonable for Debian style packaging.  The output structure is what apt actually expects, so that is not under our control.  We should adhere to the package format style (using the codename/arch spellings).  The only thing I would have done differently is split out the typo corrections into a pre-commit that could've gone in without any conversation.

stevapple:
> The only thing I would have done differently is split out the typo corrections into a pre-commit that could've gone in without any conversation.

Split out #51 √

shahmishal:
looks good to me!

futurejones:
The comments about arch type `amd64`, `arm64` and os_version `focal` not `20.04` naming for deb packages are correct. 

Naming the `package structure/url` as in `repo/ubuntu/pool/main/s/swiftlang/` etc is unnecessary as these directories and other will be automatically created when the actual repository is set up.
Package files will never be manually added to these directories in the repository. They will be added using the standard debian repository setup and maintenance tools.

A standard apt repository contains the following directories - `conf`, `db`, `dists`, `lists` and `pool`. The only directory that needs to be manually created is `conf` which contains repository `distributions` file.

If we are just talking about the location of packages for manual download on `https://download.swift.org/` then we should not be using `/pool/main/s/swiftlang/`.
A simpler location such as `https://download.swift.org/ubuntu/focal/<package-name>` would be more appropriate.
or 
`https://download.swift.org/ubuntu/focal/releases/amd64/<package-name>`,
`https://download.swift.org/ubuntu/focal/releases/aarch64/<package-name>`,
`https://download.swift.org/ubuntu/focal/dev/amd64/<package-name>`,
`https://download.swift.org/ubuntu/focal/dev/aarch64/<package-name>`,
 if you want to separate arch types and  release / snapshots etc.

`.list` files such as `swiftlang.list` are not repository configuration files. They do not live in the `dists` directory as suggested here `repo/ubuntu/dists/focal/swiftlang.list`. The correct location for `.list` files is in the `/etc/apt/sources.d/` directory.

The creation and installation of a `.list` file would normally be done by the repository installation script. The content of the `.list` file is just a `web url` and is dependent on the system OS_version and which repository section they wish to access.

compnerd:
I would say that the arch should be something to keep in place and the snapshots.  There is something to be said for the simplicity of having the results be directly hostable, but I am not tied to it.  For the URL, I suspect that path rewriting might be an option.

PR #57, opened 2021-11-16
libsqlite is pre-installed on Amazon Linux 2, removing it from requires

shahmishal:


tomerd (approved):

compnerd:
@tomerd I think that this isn't the right way to handle this.  While it may be pre-installed, the package should include it in the dependencies so that the package manager has the correct reference count on the dependency.

tomerd:
@compnerd we agree in principle. the issue here is documented in https://forums.swift.org/t/run-swift-on-al2-device/47494 which seems to be a bug in AL2 which has it pre-installed but dont recognize it as a package

futurejones:
@tomerd @compnerd @shahmishal 
The main issue here is that the package `libsqlite` is not available in amazon-linux 2. 
This is not a bug, the package has a different name, it is called `libsqlite3x`.
`sqlite` is installed by default on AWS amazon-linux 2 instances and also in the amazon-linux 2 docker images and `sqlite` includes the package `libsqlite3x`.
If you want to keep libsqlite as a dependency then the correct package is `libsqlite3x`.

tomerd:
thanks @futurejones - @shahmishal I suggest we add it back with the right name. we should also update our docker image accordingly

bmoffatt:
`sqlite` would be the right package name, it's this package that provides the `.so`

```
$ docker run public.ecr.aws/amazonlinux/amazonlinux:2 rpm -qal sqlite
/usr/bin/sqlite3
/usr/lib64/libsqlite3.so.0
/usr/lib64/libsqlite3.so.0.8.6
/usr/share/doc/sqlite-3.7.17
/usr/share/doc/sqlite-3.7.17/README
/usr/share/man/man1/sqlite3.1.gz
```

tomerd:
@shahmishal ^^

futurejones:
@tomerd , @bmoffatt is correct the correct package should be `sqlite`.
I was looking for a replacement for the missing `libsqlite` package. The only available libsqlite package is `libsqlite3x`.
But this is not needed as `sqlite` already includes `libsqlite`.

shahmishal:
https://github.com/apple/swift-installer-scripts/pull/60

PR #63, opened 2021-11-29
Update the swift version to 5.5.1

shahmishal:


platforms/Linux/RPM/Fedora/rawhide/swiftlang.spec
```diff
@@ -9,7 +9,7 @@
 %global ninja_version 1.10.2
 
 Name:           swiftlang
-Version:        5.5
+Version:        5.5.1
```
tomerd, line 12: @shahmishal could we consolidate t6he fedora setup to be like the rest of the RPMs so that we dont need to repeat this everywhere?

shahmishal: Yes, we should I will look into this.

PR #70, opened 2022-01-31
Windows: shuffle files in the MSI packaging

compnerd:
This shuffles the swift-driver from devtools to toolchain, replacing the
driver from the swift repository at packaging time.  This is required to
be this way to support bootstrapping and more importantly, to reduce the
overall build times for the toolchain (OSS builders are perilously close
to the time limit, and would fail to build with swift-driver in the same
build).  The toolchain MSI is meant to contain everything that is needed
for building a program manually with the compiler.  This should allow us
to migrate to swift-driver as the official Windows driver.

### DIFF
--- a/platforms/Windows/devtools.wxs
+++ b/platforms/Windows/devtools.wxs
@@ -67,23 +67,6 @@
       </Component>
       <?endif?>
 
-      <!-- llbuild -->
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

shahmishal (approved):

PR #72, opened 2022-02-23
Windows: add SwiftSyntaxCDataTypes.h to the toolchain

compnerd:
This repairs the toolchain manifest after apple/swift#39441 which split
the header into two files.  This ensures that both halves are packaged.


compnerd:
CC: @ahoppen

shahmishal (approved):

ahoppen:
Sounds reasonable to me.

PR #77, opened 2022-02-28
Add support for Swift version in the dir, and bump the version to Swi…

shahmishal:
…ft 5.5.3


shahmishal:
We still need to migrate Fedora over to using metadata file and docker compose.

PR #80, opened 2022-03-10
Update the RPM spec files to support release/5.6

shahmishal:


shahmishal:
cc: @tachoknight @tomerd @futurejones 

As we are getting closer to releasing Swift 5.6, I updated the spec files to include new repos and drop patch file. 

@tachoknight Did you get a chance to look into using metadata file for Fedora?

platforms/Linux/RPM/Fedora/34/swiftlang.spec
```diff
@@ -1,41 +1,61 @@
 %global debug_package %{nil}
 %global linux_version fedora
-%global swifttag 5.5.3-RELEASE
-%global swiftbuild swift-source
+%global swift_version 5.6-RELEASE
```
tomerd, line 3: should we name these 5.6.0? this could be better IMO since otherwise people will install 5.6 and not get updates?

platforms/Linux/RPM/metadata.inc
```diff
@@ -1,20 +1,26 @@
 # package metadata
 
 %global package_name      swiftlang
-%global package_version   5.5.3
+%global package_version   5.6
```
tomerd, line 4: see above re. 5.6.0 vs 5.6

tomerd (commented):
cool. one question / discussion point re. versioning of .0 releases

platforms/Linux/RPM/Fedora/34/swiftlang.spec
shahmishal: For Docker we are already tagging 5.6.0, we can do the same here.

shahmishal: I will update `%global package_version   5.6` to `%global package_version   5.6.0` and in the future we can update the swift tag scheme to include `.0`

shahmishal: Updated to use 5.6.0 package_version.

platforms/Linux/RPM/Fedora/34/swiftlang.spec
```diff
@@ -1,41 +1,62 @@
...
+%global swift_atomics_version 1.0.2
+%global swift_collections_version 1.0.1
+%global swift_numerics_version 1.0.1
+%global swift_system_version 1.1.1
+%global swift_nio_version 2.31.2
+%global swift_nio_ssl_version 2.15.0
 
 Name:           swiftlang
-Version:        5.5.3
+Version:        5.6.0
```
shahmishal, line 20:
```suggestion
Version:        %{package_version}
```

tomerd (approved):

tomerd:
nice 🚀

tachoknight:
I tried to build on Fedora 35 with the spec file but there's a couple of issues, one is minor and I created [this PR](https://github.com/apple/swift-installer-scripts/pull/84) to fix it. The build doesn't currently work with the latest tag (5.6-DEVELOPMENT-SNAPSHOT-2022-03-02-a) so I guess I'll wait till 5.6 is released to make sure it builds.

shahmishal:
> I tried to build on Fedora 35 with the spec file but there's a couple of issues, one is minor and I created [this PR](https://github.com/apple/swift-installer-scripts/pull/84) to fix it. The build doesn't currently work with the latest tag (5.6-DEVELOPMENT-SNAPSHOT-2022-03-02-a) so I guess I'll wait till 5.6 is released to make sure it builds.

Thank you for the fix!

PR #81, opened 2022-03-11
update RPM specs dependencies

tomerd:
motivation: static linking required the development version of libxml2 and libcurl

changes: include libxml2-devel and libcurl-devel which are needed for static linking


tomerd:
@shahmishal @drexin

shahmishal (approved):

PR #91, opened 2022-03-20
windows: update toolchain manifest for swift-argument-parser

compnerd:
Enable statically linking the swift-argument-parser library
SwiftArgumentParserToolInfo into SwiftAgumentParser as there is only a
single user of the library.


compnerd:
This will require some additional coordination before merging: we need a new tagged version of apple/swift-argument-parser with apple/swift-argument-parser#424 and then we need to bump the update-checkout-config.json in apple/swift.

shahmishal (approved):

PR #97, opened 2022-03-29
WiX: only setup SDKROOT for native platform

compnerd:


platforms/Windows/sdk.wixproj
```diff
@@ -56,6 +56,10 @@
   </Target>
 
   <ItemGroup>
+    <WixExtension Include="WixUtilExtension">
```
wjk, line 59: This doesn’t look like it is used.

compnerd: IIRC, I needed it for the platform computation.

compnerd:
While this would be nice, the SDKs are co-installed to the same location, so I'm just going to close this for now.

PR #98, opened 2022-03-29
WiX: restructure XCTest in SDK

compnerd:
The rest of the Windows setup is properly configured for multiple
parallel SDKs and versions.  However, the XCTest import library (which
would not be a problem if all SDKs were guaranteed to be installed and
we did some pre-processing before building the MSIs) and more
importantly the runtime DLL for all architectures is in a singular
location.  This would be a problem for installing multiple SDKs for the
same platform but multiple architectures.  Now that we are getting
closer to the multi-architecture cross-compiling details being
finalized, we need to repair this oversight.

This breaks the layout and compatibility with older toolchains.
However, it is something that the user _could_ workaround if they are
sufficiently motivated (it requires additional flags to be explicitly
specified).  An associated change to SPM enables this use of layout.


compnerd:
This is associated with apple/swift-package-manager#4266.  That is merged so this should be compatible with a SPM building whenever this gets merged.

shahmishal (approved):

PR #119, opened 2022-04-18
WiX: remove extraneous architecture in search path

compnerd:
Adjust the location of the swift modules to overlap across SDKs.  The
directory form of the swift module can easily overlap and co-exist.  Now
that the driver has learnt to scan the non-architecture dependent location
in addition to the architecture location dependent path, we can unify
the swiftmodule content in the SDK to a simpler layout.  The layout now
is:

~~~
SDKROOT
  `- usr
      `- lib
          `- swift
              `- [platform]
                  +- [module].swiftmodule
                  |   `- [triple].{swiftdoc,swiftmodule,swiftinterface}
                  `- [arch]
                      `- [import library]
~~~

On Windows, it would be preferable to further remove the intermediate
platform as we have that information earlier in the path.

Subsequent work remains to now ensure that the package manager will
similarly allow the XCTest swift modules to collapse.

While in theory it is possible to create a multi-architecture import
library, it is valuable to support user-selectable SDK installation
which requires that the we create the import libraries on the fly for
the selected architectures.  It is more convenient from a distribution
perspective to simply make the content static and as such continue to
use the architect dependent directory for the import library unlike the
swift module.

This is dependent on apple/swift#42419.

### DIFF
--- a/platforms/Windows/sdk-amd64.wxs
+++ b/platforms/Windows/sdk-amd64.wxs
@@ -62,33 +62,33 @@
                           <Directory Id="WindowsSDK_usr_lib_swift_shims" Name="shims">
                           </Directory>
                           <Directory Id="WindowsSDK_usr_lib_swift_windows" Name="windows">
+                            <Directory Id="_Concurrency.swiftmodule" Name="_Concurrency.swiftmodule">
+                            </Directory>
+                            <Directory Id="_Differentiation.swiftmodule" Name="_Differentiation.swiftmodule">
+                            </Directory>
+                            <Directory Id="Distributed.swiftmodule" Name="Distributed.swiftmodule">
+                            </Directory>
+                            <Directory Id="_RegexParser.swiftmodule" Name="_RegexParser.swiftmodule">
+                            </Directory>
+                            <Directory Id="_StringProcessing.swiftmodule" Name="_StringProcessing.swiftmodule">
+                            </Directory>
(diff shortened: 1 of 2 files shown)


stevapple:
While I’m delighted by the change, we had better first have it reflected in the CMake scripts for core libraries.

Unix (Linux) always supports multi-arch Swift library bundle, and I’ve taken a try on apple/swift#41850 to align the runtime library layout of Windows and Unix. This will further enable us to have portable and multi-arch SDKs with unified layout.

Discussion thread: https://forums.swift.org/t/misplacement-of-runtime-libraries-on-linux/55888

compnerd:
The CMake build already does this correctly, this is undoing the need for the workaround of moving the content around at packaging time.

stevapple:
> Edit: Wait, do you mean in swift-corelibs-{libdispatch,foundation,xctest}?

Exactly.

compnerd:
What do you think would be good to change there?  I don't mind doing that subsequently (this change is already complicated enough - its across 3 different repositories).

stevapple:
It’s nice to hear that’s going to be addressed. Sequence is not so important, and go ahead 🚀

compnerd:
Ah, I see.  Yes, I agree, it would be nice if CMake installed the files in the final location.  I can do those changes subsequently.  However, longer term, I want to get rid of that and use `install(TARGETS ...)` with CMake handling the layout.

shahmishal (approved):

PR #120, opened 2022-04-26
WiX: remove extraneous architecture in search path

compnerd:
Adjust the location of the swift modules to overlap across SDKs.  The
directory form of the swift module can easily overlap and co-exist.  Now
that the driver has learnt to scan the non-architecture dependent location
in addition to the architecture location dependent path, we can unify
the swiftmodule content in the SDK to a simpler layout.  The layout now
is:

~~~
SDKROOT
  `- usr
      `- lib
          `- swift
              `- [platform]
                  +- [module].swiftmodule
                  |   `- [triple].{swiftdoc,swiftmodule,swiftinterface}
                  `- [arch]
                      `- [import library]
~~~

On Windows, it would be preferable to further remove the intermediate
platform as we have that information earlier in the path.

Subsequent work remains to now ensure that the package manager will
similarly allow the XCTest swift modules to collapse.

While in theory it is possible to create a multi-architecture import
library, it is valuable to support user-selectable SDK installation
which requires that the we create the import libraries on the fly for
the selected architectures.  It is more convenient from a distribution
perspective to simply make the content static and as such continue to
use the architect dependent directory for the import library unlike the
swift module.

This is dependent on apple/swift#42419.

### DIFF
--- a/platforms/Windows/sdk-amd64.wxs
+++ b/platforms/Windows/sdk-amd64.wxs
@@ -62,33 +62,33 @@
                           <Directory Id="WindowsSDK_usr_lib_swift_shims" Name="shims">
                           </Directory>
                           <Directory Id="WindowsSDK_usr_lib_swift_windows" Name="windows">
+                            <Directory Id="_Concurrency.swiftmodule" Name="_Concurrency.swiftmodule">
+                            </Directory>
+                            <Directory Id="_Differentiation.swiftmodule" Name="_Differentiation.swiftmodule">
+                            </Directory>
+                            <Directory Id="Distributed.swiftmodule" Name="Distributed.swiftmodule">
+                            </Directory>
+                            <Directory Id="_RegexParser.swiftmodule" Name="_RegexParser.swiftmodule">
+                            </Directory>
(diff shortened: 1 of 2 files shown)


shahmishal (approved):

shahmishal:
@compnerd is this ready to be merged?

compnerd:
Yes, I believe it is; I really wish that we could kick off toolchain builds that we could easily test with the changes.

shahmishal:
You can using the cross PR testing

compnerd:
Hmm, it seems that it still doesn't report the generated installer - is there a way to get that to test?

shahmishal:
> Hmm, it seems that it still doesn't report the generated installer - is there a way to get that to test?

You should be able to download it now. I started a build over the weekend.

compnerd:
Bleh, seems to be missing a change in swift-driver; thanks for setting up the download!

compnerd:
Just verified, this is ready to go.  Thanks for the ping on this @shahmishal!  This is actually going to help make the 5.7 release for Windows even better :)

PR #133, opened 2022-06-16
Update the version to Swift 5.6.2

shahmishal:


tomerd (approved):

tomerd:
we need to clean up this patches from deb, so we don't have to manually do all these version updates in a non-centralized way

shahmishal:
Agree, @jblache has PRs open to remove these in the future.

PR #137, opened 2022-07-20
WiX: enable selection of installation path

lxbndr:
This adds a simple target folder selection UI to Windows MSI bundles.

Everything is made as simple as possible with the WixUI extension.


lxbndr:
CC @compnerd

platforms/Windows/installer.wxs
```diff
@@ -2,14 +2,17 @@
...
     </BootstrapperApplicationRef>
 
+    <Variable Name="InstallFolder" Type="string" Value="[WindowsVolume]Library"/>
+
     <Chain>
       <?if $(var.RequiredChain) != "" ?>
       <?foreach MSI in $(var.RequiredChain) ?>
       <MsiPackage SourceFile="$(var.MSI_LOCATION)\$(var.MSI)" Compressed="yes">
         <MsiProperty Name="INSTALL_DEBUGINFO" Value="[INSTALL_DEBUGINFO]" />
+        <MsiProperty Name="INSTALLDIR" Value="[InstallFolder]" />
```
compnerd, line 15: Because this unconditionally sets this for all MSIs, does this now not change the default install location for the runtime and ICU?  Both of these have moved to `%ProgramFiles%\swift` in 5.7, so this might actually move them back out?

compnerd: Yeah, I think that we want to start migrating the runtimes to the appropriate places.  See https://github.com/apple/swift-installer-scripts/commit/681954743ec2f83f3d5020c3bece5d53936623d3

This is part of why I think that we might want to hold off on the option being exposed in the installer but allowing exposing it in the MSIs.

lxbndr: I see... Looks like we need some advanced path setup, so a user could select different paths for installed features. I've reverted the installer. Now this change is about MSI only.

lxbndr: Updated the PR description

compnerd (approved):
Should verify the installer, but the changes look good to me.

compnerd:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/140/artifact/build/artifacts/installer.exe

compnerd (approved):

shahmishal (approved):

airaolagoitia:
Would it be possible to set the installation path as an argument to the executable? I'm thinking on automating the toolchain installation (so no user intervention) and being able to set it through CLI.

compnerd:
There’s more work that needs to be done in order to support that.  Patches to incrementally improve the installer to support that are welcome.  But eventually it should be.

PR #139, opened 2022-08-20
WiX: Improvements to Windows packaging

stevapple:
- Update Program Files entry name to `Swift`, as respecting Windows convention.
- Add missing `Cxx` library in SDK and runtime profiles.

### DIFF
--- a/platforms/Windows/runtime-amd64.wxs
+++ b/platforms/Windows/runtime-amd64.wxs
@@ -57,6 +57,9 @@
       <Component Id="swift_StringProcessing.dll" Guid="198f58ad-8797-40ee-b109-dec6f9b57b27">
         <File Id="swift_StringProcessing.dll" Source="$(var.SDK_ROOT)\usr\bin\swift_StringProcessing.dll" Checksum="yes" />
       </Component>
+      <Component Id="swiftCxx.dll" Guid="f5af3d1d-bb5f-4fae-b61c-ade58a581aec">
+        <File Id="swiftCxx.dll" Source="$(var.SDK_ROOT)\usr\bin\swiftCxx.dll" Checksum="yes" />
+      </Component>
       <Component Id="swiftCore.dll" Guid="4098dff8-8b8d-48ee-a234-d29104d4c809">
         <File Id="swiftCore.dll" Source="$(var.SDK_ROOT)\usr\bin\swiftCore.dll" Checksum="yes" />
       </Component>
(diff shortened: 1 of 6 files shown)


stevapple:
cc @egorzhdan , did you add anything besides the `Cxx` module for C++ interop?

platforms/Windows/runtime-amd64.wxs
```diff
@@ -21,7 +21,7 @@
       </Directory>
     </Directory>
 
-    <SetDirectory Id="INSTALLDIR" Value="[ProgramFiles64Folder]swift">
+    <SetDirectory Id="INSTALLDIR" Value="[ProgramFiles64Folder]Swift">
```
compnerd, line 24: This was intentionally `swift`.  I don't think that there is any strict suggestion/requirement/pattern that says that this should be `Swift`.

platforms/Windows/runtime-arm64.wxs
```diff
@@ -21,7 +21,7 @@
       </Directory>
     </Directory>
 
-    <SetDirectory Id="INSTALLDIR" Value="[ProgramFiles64Folder]swift">
+    <SetDirectory Id="INSTALLDIR" Value="[ProgramFiles64Folder]Swift">
```
compnerd, line 24: Similar

compnerd (changes requested):
Nice catch on the new module!

platforms/Windows/runtime-amd64.wxs
stevapple: MSDN suggests the installation path to be `[Program64FilesFolder]\ISV Name\Application Name\x64` (take x86_64 as example), and it’s clear to use title case with respect to blank spaces.

In practice, `ISV Name` is sometimes ignored, arch directory is ignored if it doesn’t support multi-arch installation, major version number is appended after `Application Name` if it supports multi-version installation.

This means if possible we should use `[Program64FilesFolder]\Swift\5.8-dev\runtime\x64` or `[Program64FilesFolder]\Swift\5.8-dev\runtime`.

stevapple: On my machine this rule gets fully addressed except for `dotnet` and `nodejs`. I don’t think we should be the rule breaker if we can avoid that.

image

egorzhdan:
> cc @egorzhdan , did you add anything besides the Cxx module for C++ interop?

I don't recall adding something else, other than the C++ stdlib overlay (`swiftstd`) and the `CxxShim` module, both of which don't currently work on Windows.

platforms/Windows/runtime-amd64.wxs
stevapple: "ISV Name" uses title case, and I believe you’re confusing the "Application Name" here. An application on Windows can contain several components.

For Swift, "ISV Name" should be `Swift` or `Swift.org`, and "Application Name" should be `Swift`.

stevapple: `Swift` looks more fit and I don’t think we have reason to refuse that🤷‍♂️ Other OSS projects like Git, CMake and Vim also respects title case on Windows — when in Rome, do as Romans do.

Also, according to Swift editorial guideline:
> When using the name Swift in titles, headlines, or body copy, always typeset Swift with an uppercase S followed by lowercase letters.

We should prefer `Swift` to `swift` where title case is allowed.

compnerd:
@stevapple could you rebase this change please?

PR #140, opened 2022-08-24
WiX: package up clangd on Windows

compnerd:
We were not packaging clangd.exe into the toolchain distribution which
is required for SourceKit-LSP.  Add the binary into the ARM64 and AMD64
packages.


shahmishal (approved):

compnerd:
This does add some bulk to the installer, however, this is also needed to enable some of the SourceKit-LSP improvements.  This seems to improve the LSP support and helps some of the test suite in the LSP.  I think that given the dependency for LSP, it makes sense to package it up even in light of the additional size implications.

PR #164, opened 2023-01-15
WiX: Move `WinSDK.modulemap` into SDK

stevapple:
Update WiX files for apple/swift#63042


compnerd:
The custom action is now completely removed, I think that this is safe to close.  If there is anything remaining that should be picked up from this, please rebase and reopen the PR.

PR #166, opened 2023-01-19
Add DependsOnTargets to SignBundle insignia target

amonshiz:
While looking through this file I noticed that only the "SignBundle" target was missing a DependsOnTarget clause. All the others had it so seems reasonable for "SignBundle" to also.


compnerd (approved):
Thanks!

shahmishal (approved):

PR #174, opened 2023-02-08
add dependencies on zip and unzip where needed

tomerd:


shahmishal (approved):

compnerd (approved):
Seems reasonable, but please change the title - it doesn't add it everywhere, it adds it on Linux platforms.  Where is the packaging change for Windows? ;-)

stevapple:
For Windows it seems more reasonable to add them as WinGet dependencies? If this is planned for Swift 5.9, we should remember to do that when it’s released.

tomerd:
I think on windows we use tar?

stevapple:
I think that’s a workaround introduced by @compnerd? `tar.exe` is basically the same program on Linux. It doesn’t support `.zip` archives. We were using it because there’s no `zip` and `unzip` installed on Swift CI. This may be something to add first.

compnerd:
@stevapple - correct, we use `tar`, but it does support zip! (At least it did as of ~2 days ago when I tried?). The point was that this is Linux specific, not "everywhere".

stevapple:
I tried, and that’s surprising @compnerd — yet another undocumented Microsoft magic…

You’re right that we should limit the scope to "on Linux":)

compnerd:
@stevapple its actually less surprising and more obvious in hindsight.  The most likely source for this is BSD zip, which IIRC supports libarchive as a backend, which supports zip.

PR #177, opened 2023-03-07
Move installed XCTest.swiftmodule up one directory on Windows

tristanlabelle:
The layout of the XCTest directories on Windows is currently inconsistent between local builds and installed builds. This change resolves the inconsistency by updating swiftmodule directory location for XCTest from:

`XCTest\usr\lib\swift\windows\%ARCH%\XCTest.swiftmodule`
To
`XCTest\usr\lib\swift\windows\XCTest.swiftmodule`

This allows us to use the same include path on all architectures and matches other standard modules. SPM was previously updated accordingly here: https://github.com/apple/swift-package-manager/pull/4300 but the related #119 did not update the paths for `XCTest`.


compnerd (changes requested):
This breaks the installer, please test the change locally.  The directory structure is needed as the import library will be installed there.

compnerd (approved):
Need to test this, but the change itself should be good to do.

PR #180, opened 2023-03-15
Allow overriding the XCTest directory in the Windows installer build

tristanlabelle:
The paths where the Windows installer expects to finds XCTest binaries would conflict between architectures. This msbuild property allows us to have parallel XCTest cmake install trees so the `usr\lib\swift\windows\XCTest.lib` path does not collide.

Related changes:
- https://github.com/apple/swift/pull/64425
- https://github.com/compnerd/swift-build/pull/480

### DIFF
--- a/platforms/Windows/sdk-amd64.wxs
+++ b/platforms/Windows/sdk-amd64.wxs
@@ -118,23 +118,23 @@
     <!-- Components -->
     <ComponentGroup Id="XCTest">
       <Component Id="XCTest.dll" Directory="XCTest_usr_bin64" Guid="11b12995-a7e6-4de3-954d-d4ddc977633f">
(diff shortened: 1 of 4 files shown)


compnerd:
This should be associated with a change to apple/swift utils/build-windows-toolchain.bat

tristanlabelle:
@compnerd The change is backwards compatible. If XCTEST_ROOT is not set explicitly, it will default to what it previously was hard-coded to:

```xml
<XCTEST_ROOT Condition=" '$(XCTEST_ROOT)' == '' ">$(PLATFORM_ROOT)\Developer\Library\XCTest-development</XCTEST_ROOT>

compnerd:
@tristanlabelle why not just update the three places that use this?
apple/swift:utils/build-windows-toolchain.bat
compnerd/swift-build:.github/workflows/swift-toolchain.yml
compnerd/swift-build:.azure/vs2022.yml

tristanlabelle:
@compnerd I followed your suggestion and added the related changes. Would you then remove the default value for `XCTEST_ROOT`?

PR #181, opened 2023-03-15
Expect XCTest.lib in arch-disambiguated directory in the Windows installer

tristanlabelle:
Updates the Windows installer to account for https://github.com/apple/swift-corelibs-xctest/pull/434


compnerd:
I'm not sure how this worked previously, but ... okay.

tristanlabelle:
It wouldn't have worked if building for two different architectures and then the msi's for those two architectures (they would both have picked up the latter lib file), but it would have worked if building one architecture, its msi, then the other architecture, and then its msi.

compnerd:
I think that we staged to different directories, I just don't understand how I never ran into any issues yet 🤯

tristanlabelle:
Abandonning in favor of handling this in the build scripts.

PR #185, opened 2023-03-30
Fix default Windows install directory regression from WiX v4 port

tristanlabelle:
The rules around `<SetDirectory>` changed between WiX v3 and v4, so our port resulted in the INSTALLDIR defaulting to `C:\` instead of either `C:\Program Files\swift` or `C:\Library`.

Changes:
- For the runtime, use `<StandardDirectory>` to root in `ProgramFiles`, and capitalize `Swift` for consistency with all other programs
- For stuff installed under `C:\Library`, use `<SetDirectory>` + `<Directory>` as a means of rooting at `WindowsVolume`, also change the customizable installation path to the `C:\` underneath `Library` (so the installer will write the `Library` directory under the users' path rather than start at the `Developer` directory)

How tested:
- Rebuilt and installed runtime + sdk msi's to their default locations and to overriden locations, then installed the full installer. Verified that paths and environment variables are as expected in all cases.


tristanlabelle:
Oops there seems to be a bad merge in there, gimme a sec

tristanlabelle:
False alarm, there's no bad merge, just a confusing diff.

compnerd (approved):

compnerd:
Tested at apple/swift#64815

PR #188, opened 2023-04-13
WiX: permit building the installer for ARM64

compnerd:
Windows ARM64 support is slowly being bootstrapped, enable the ARM64 installer which differs only in the upgrade code and the architecture name.


platforms/Windows/installer-arm64.wxs
```diff
@@ -0,0 +1,25 @@
+﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:bal="http://wixtoolset.org/schemas/v4/wxs/bal">
+  <Bundle Name="Swift Developer Package for Windows aarch64" Version="$(var.ProductVersion)" Manufacturer="swift.org" UpgradeCode="151d42d9-8877-4b72-ac62-4695243a35c1">
```
tristanlabelle, line 2: I think the `UpgradeCode` is the one Guid that we care about being unique per platform.

tristanlabelle (approved):
Check the `UpgradeCode` comment but otherwise this looks good

platforms/Windows/installer-arm64.wxs
compnerd: Yeah, it is unique in this case.

tristanlabelle: Oh my bad, I thought I had seen it being the same as amd64

compnerd: FTR (and because you scared me and made me check it again!) The X64 installer upgrade code is 8c75f32a-7bdf-4c61-abf6-c7e1c4b8fbf6.

shahmishal, line 2: Do we want to call it `aarch64` or `arm64`?

compnerd: I do want to rename it to ARM64, but I would rather do that in a single pass (I have exactly that change queued up :)).  This maintains the current spelling of aarch64 which is used in the MSIs.  I am going to be doing a much more invasive renaming when I do the ARM64 rename because this does get shown in a few places.

shahmishal (approved):

PR #200, opened 2023-05-26
WiX: package up CompilerPluginSupport module

compnerd:
This is part of the Manifest API in SPM and expected to be present. Package this into the distribution as well.


etcwilde (approved):
Sure

PR #201, opened 2023-05-26
WiX: version the installations of the toolchain, devtools, runtime

compnerd:
These packages are meant to be parallel installed with different releases except in the case of development releases.  Wire up the package version information into the path.


platforms/Windows/runtime-amd64.wxs
```diff
@@ -13,8 +13,12 @@
     <!-- Directory Structure -->
     <StandardDirectory Id="ProgramFiles64Folder">
       <Directory Id="INSTALLDIR" Name="Swift">
-        <!-- TODO(compnerd) use $(var.ProductVersion) -->
-        <Directory Id="_" Name="runtime-development">
+        <?if $(var.ProductVersion) = "0.0.0"?>
```
tristanlabelle, line 16: If this isn't scoping dependent and is referenced later in the file, shouldn't we pull it to the top level and possibly at the top of the file? (same for other files)

etcwilde (approved):
Verified base installer still works. LGTM

Note that changing the name of the toolchain might break some existing tools where it's hard-coded.

compnerd:
Yeah, there is no good way around that.  If they hardcoded the values, they knew this was likely to happen.

PR #202, opened 2023-05-27
WiX: simplify the swift-format package definition

compnerd:
- Use durable GUID generation for auto-generation of the GUID.
- Combine the platform definitions into a singular source to avoid drift.

This avoids having to keep the variants in sync and provides an easier upgrade path.  This is the first step towards enabling a non-admin experience for users.


tristanlabelle (approved):
Nice!

etcwilde (approved):
install seems to work with swiftc. Can't really test the swift-format tool since it's not included in the build, so this control path wasn't really doing anything at the moment anyway.

PR #204, opened 2023-06-02
Add wix files for swift-inspect.

hjyamauchi:


platforms/Windows/swift-inspect.wixproj
```diff
@@ -0,0 +1,32 @@
...
+    <ProductVersion Condition=" '$(ProductVersion)' == '' ">0.0.0</ProductVersion>
+    <ProductVersion>$(ProductVersion)</ProductVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <OutputPath>build\</OutputPath>
+    <IntermediateOutputPath>build\obj\</IntermediateOutputPath>
+  </PropertyGroup>
+
+  <!-- <Import Project="WiXCodeSigning.targets" /> -->
```
compnerd, line 19: I think that we should be able to restore this, the targets are in the same location.

hjyamauchi: Done.

compnerd (approved):

PR #205, opened 2023-06-02
WiX: simplify the toolchain package definition

compnerd:
- Use durable GUID generation for auto-generation of the GUID.
- Combine the platform definitions into a singular source to avoid drift.

This avoids having to keep the variants in sync and provides an easier upgrade path. This is the first step towards enabling a non-admin experience for users.


etcwilde (approved):
Yeah, looks like everything is in there.

PR #210, opened 2023-06-08
WiX: package RegexBuilder into the SDK on Windows

compnerd:
We were previously missing the RegexBuilder module in the SDK distribution.  Package this up as it should be available on Windows.


DougGregor (approved):

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/1012/

PR #211, opened 2023-06-16
WiX: clean up the tools packaging for Windows

compnerd:
Clean up the build rules for swift-inspect and swift-format to prepare for the packaging of the tools in the default distribution.

### DIFF
--- a/platforms/Windows/swift-format.wixproj
+++ b/platforms/Windows/swift-format.wixproj
@@ -19,7 +19,22 @@
   <Import Project="WiXCodeSigning.targets" />
 
   <PropertyGroup>
-    <DefineConstants>ProductVersion=$(ProductVersion);SWIFT_FORMAT_BUILD=$(SWIFT_FORMAT_BUILD)</DefineConstants>
+    <DefineConstants>ProductArchitecture=$(ProductArchitecture);ProductVersion=$(ProductVersion);SWIFT_FORMAT_BUILD=$(SWIFT_FORMAT_BUILD)</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'amd64' ">
+    <InstallerPlatform>x64</InstallerPlatform>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm64' ">
+    <InstallerPlatform>arm64</InstallerPlatform>
+    <InstallerVersion>500</InstallerVersion>
+  </PropertyGroup>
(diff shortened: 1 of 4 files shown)


platforms/Windows/swift-format.wixproj
```diff
@@ -19,7 +19,22 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm64' ">
+    <InstallerPlatform>arm64</InstallerPlatform>
+    <InstallerVersion>500</InstallerVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm' ">
+    <InstallerPlatform>arm</InstallerPlatform>
+    <InstallerVersion>500</InstallerVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'x86' ">
```
etcwilde, line 36: The `ProgramFiles64Folder` base directory in swift-format.wxs will probably break if we ever try to actually install to a 32-bit system, no?

platforms/Windows/swift-inspect.wixproj
```diff
@@ -19,7 +19,22 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm64' ">
+    <InstallerPlatform>arm64</InstallerPlatform>
+    <InstallerVersion>500</InstallerVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm' ">
+    <InstallerPlatform>arm</InstallerPlatform>
+    <InstallerVersion>500</InstallerVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'x86' ">
```
etcwilde, line 36: Same comment, 32-bit probably doesn't work.

etcwilde (commented):
32-bit platforms seem slightly broken. The WIXproject suggests it works on 32-bit windows, but the install directory structure won't work.

platforms/Windows/swift-format.wixproj
compnerd: Yes, but Windows only supports 64-bit development hosts.  Do we want to treat `swift-format` as a client application or a toolchain tool? 🤔

platforms/Windows/swift-inspect.wixproj
compnerd: Similar.

PR #213, opened 2023-06-22
[cxx-interop] CxxShim files became arch-independent

egorzhdan:
This adjusts the Windows installation logic after the changes in https://github.com/apple/swift/pull/66765

rdar://110788977

### DIFF
--- a/platforms/Windows/sdk.wxs
+++ b/platforms/Windows/sdk.wxs
@@ -416,12 +416,12 @@
     </ComponentGroup>
 
     <ComponentGroup Id="libcxxshim">
-      <Component Directory="WindowsSDK_usr_lib_swift_windows_ARCH" Id="libcxxshim.h">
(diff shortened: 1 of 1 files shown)


platforms/Windows/sdk.wxs
```diff
@@ -417,11 +417,11 @@
 
     <ComponentGroup Id="libcxxshim">
       <Component Directory="WindowsSDK_usr_lib_swift_windows_ARCH" Id="libcxxshim.h">
```
compnerd, line 419: This needs to change as well, as while the point where it is installed to was updated, the `Directory` attribute here controls where it is deployed to.

```suggestion
      <Component Directory="WindowsSDK_usr_lib_swift_windows" Id="libcxxshim.h">
```

compnerd (changes requested):
Thanks!

platforms/Windows/sdk.wxs
egorzhdan: Fixed, thanks!

compnerd (approved):
Probably a good idea to run a build.

PR #215, opened 2023-07-25
Refactor WiX installer projects

barnson:


platforms/Windows/bld/bld.wixproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="WixToolset.Sdk/4.0.1">
+  <PropertyGroup>
+    <DefineConstants>
+      $(DefineConstants);
+      DEVTOOLS_ROOT=$(DEVTOOLS_ROOT);
+      TOOLCHAIN_ROOT=$(TOOLCHAIN_ROOT);
+      TOOLCHAIN_ROOT_USR_LIB_CLANG=$(TOOLCHAIN_ROOT_USR_LIB_CLANG);
```
compnerd, line 7: Not sure if there is a better way to do this, IIRC this was required for the use of heat.

platforms/Windows/bld/bld.wxs
```diff
@@ -0,0 +1,351 @@
...
+    </DirectoryRef>
+
+    <DirectoryRef Id="_usr">
+      <Directory Name="share">
+        <Directory Id="_usr_share_swift" Name="swift" />
+      </Directory>
+    </DirectoryRef>
+
+    <ComponentGroup Id="binutils" Directory="_usr_bin">
+      <!-- TODO(compnerd) can we use symbolic links to llvm-ar.exe instead? -->
```
compnerd, line 28: Would be great to know if you have an answer to this question - is there a way to use a symbolic link that can be laid down rather than duplicating the binary itself.

platforms/Windows/bld/bld.wxs
```diff
@@ -0,0 +1,351 @@
...
+      </Component>
+    </ComponentGroup>
+
+    <ComponentGroup Id="EnvironmentVariables">
+      <Component Id="SystemEnvironmentVariables" Condition="ALLUSERS=1" Directory="INSTALLDIR" Guid="d01ea5b8-0f8a-4388-9b61-1186efddfc39">
+        <Environment Id="SystemPath" Action="set" Name="Path" Part="last" Permanent="no" System="yes" Value="[INSTALLDIR]Toolchains\$(var.ProductVersion)+Asserts\usr\bin" />
+      </Component>
+      <Component Id="UserEnvironmentVariables" Condition="NOT ALLUSERS=1" Directory="INSTALLDIR" Guid="ab52b870-23ee-42e8-9581-3fcbdfb9228c">
+        <Environment Id="UserPath" Action="set" Name="Path" Part="last" Permanent="no" System="no" Value="[INSTALLDIR]Toolchains\$(var.ProductVersion)+Asserts\usr\bin" />
+      </Component>
```
compnerd, line 331: Would be good to know if we should preserve this - this is meant to support the dual-scope installers.

platforms/Windows/bundle/installer.wixproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="WixToolset.Sdk/4.0.1">
+  <PropertyGroup>
+    <OutputType>Bundle</OutputType>
+    <OutputName>swift-$(ProductVersion)-X-windows10</OutputName>
```
compnerd, line 4: Curious about the `-X-` infix here, what is the reasoning for that?

platforms/Windows/bundle/installer.wxs
```diff
@@ -0,0 +1,33 @@
...
+    <RelatedBundle Id="8c75f32a-7bdf-4c61-abf6-c7e1c4b8fbf6" Action="upgrade" />
+    <RelatedBundle Id="151d42d9-8877-4b72-ac62-4695243a35c1" Action="upgrade" />
+
+    <BootstrapperApplication>
+      <bal:WixStandardBootstrapperApplication
+      LicenseUrl="https://www.swift.org/LICENSE.txt"
+      LogoFile="swift.png"
+      SuppressOptionsUI="yes"
+      SuppressRepair="no"
+      Theme="hyperlinkLicense" />
```
compnerd, line 19: The other locations indent the continuation, would be good to maintain the style.

```suggestion
        LicenseUrl="https://www.swift.org/LICENSE.txt"
        LogoFile="swift.png"
        SuppressOptionsUI="yes"
        SuppressRepair="no"
        Theme="hyperlinkLicense" />
```

platforms/Windows/bld/bld.wxs
```diff
@@ -0,0 +1,351 @@
+﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:ui="http://wixtoolset.org/schemas/v4/wxs/ui">
+  <Package
+      Language="1033"
+      Manufacturer="swift.org"
+      Name="!(loc.Bld_ProductName)"
+      UpgradeCode="7e95dc06-7f84-4e8e-a038-8304af0468fb"
+      Version="$(var.ProductVersion)"
+      Scope="$(PackageScope)">
+
+    <MediaTemplate CabinetTemplate="bld{0}.cab" />
```
compnerd, line 10: What does the `{0}` signify/do?

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -0,0 +1,16 @@
...
+  <String Id="Ide_ProductName" Value="Swift IDE Integration Tools" />
+  <String Id="Runtime_ProductName_AMD64" Value="Swift Utilities Windows (AMD64)" />
+  <String Id="Runtime_ProductName_ARM64" Value="Swift Utilities Windows (ARM64)" />
+  <String Id="Sdk_ProductName_AMD64" Value="Swift Windows SDK (AMD64)" />
+  <String Id="Sdk_ProductName_ARM64" Value="Swift Windows SDK (ARM64)" />
+  <String Id="Sdk_ProductName_x86" Value="Swift Windows SDK (x86)" />
+  <String Id="Fmt_ProductName" Value="Swift Code Formatter" />
+  <String Id="Inspect_ProductName" Value="SwiftInspect" />
+  <String Id="BundleName" Value="Swift Software Development Kit" />
+</WixLocalization>
```
compnerd, line 16: Nit: missing newline

platforms/Windows/bundle/installer.wxs
```diff
@@ -0,0 +1,33 @@
...
+    </BootstrapperApplication>
+
+    <Chain>
+      <MsiPackage SourceFile="$(MSI_LOCATION)\runtime.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\bld.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\cli.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\dbg.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\ide.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\swift-format.msi" />
+      <MsiPackage SourceFile="$(MSI_LOCATION)\swift-inspect.msi" />
```
compnerd, line 29: Unfortunately, these are not yet generated in all builds.  If they were, I think that we would want to put `swift-format` into `cli.msi` and `swift-inspect` into `dbg.msi`.  Perhaps we can just merge the entries there with an option to just warn if the file is not present and continuing?

barnson: We can conditionalize on the architecture. Do you want me to move them now? It's just as easy to conditionalize in the bundle as in a package?

platforms/Windows/bundle/installer.wxs
```diff
@@ -0,0 +1,35 @@
+﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:bal="http://wixtoolset.org/schemas/v4/wxs/bal">
+  <Bundle
+    Name="!(loc.BundleName)"
+    Version="$(var.ProductVersion)"
+    Manufacturer="swift.org"
+    Compressed="$(IsBundleCompressed)"
+    UpgradeCode="710F1827-DA4A-4BF4-BDCE-D5F2D7C0DEF2">
```
tristanlabelle, line 7: Is it no problem to have the same UpgradeCode for both architectures?

Would that prevent installing both the arm64 and amd64 binaries on an arm64 machine (which supports amd64 emulation)? (maybe not a real scenario, asking to understand)

platforms/Windows/bundle/installer.wxs
compnerd: I think that we can just move them now as long as they don't break the build.  They are intended to be built always and so rather than increasing the number of MSIs further we can reduce them a bit and possibly reduce the amount of code for this.  The preprocessor and property should avoid the MSI for the single entry (the project references make the MSIs seem even sillier than when they are standalone).

barnson: Sounds reasonable. I'll wire that up in another commit today.

barnson: (BTW, build-windows-toolchain.bat doesn't build those packages today, so it's actually going to save the build to move them into existing packages.)

platforms/Windows/bundle/installer.wxs
barnson: You're right; either architecture would upgrade its own and the other on an Arm64 machine. If making them side-by-side is a real scenario, then they need their own upgrade codes and some directory reorganization so they don't overlap. Windows Installer supports overlap for non-PE files like SDK headers but all the host binaries would need architecture-specific directories to avoid overlap. We can figure out the directories as part of the per-user work if you want it -- especially as we're already messing around with how the tree is organized by moving it out of Program Files.

tristanlabelle: @compnerd Do you think we care about this?

tristanlabelle (approved):

platforms/Windows/bld/bld.wxs
```diff
@@ -324,10 +325,10 @@
 
     <ComponentGroup Id="EnvironmentVariables">
       <Component Id="SystemEnvironmentVariables" Condition="ALLUSERS=1" Directory="INSTALLDIR" Guid="d01ea5b8-0f8a-4388-9b61-1186efddfc39">
-        <Environment Id="SystemPath" Action="set" Name="Path" Part="last" Permanent="no" System="yes" Value="[INSTALLDIR]Toolchains\$(var.ProductVersion)+Asserts\usr\bin" />
+        <Environment Id="SystemPath" Action="set" Name="Path" Part="last" Permanent="no" System="yes" Value="[_usr_bin]" />
       </Component>
       <Component Id="UserEnvironmentVariables" Condition="NOT ALLUSERS=1" Directory="INSTALLDIR" Guid="ab52b870-23ee-42e8-9581-3fcbdfb9228c">
-        <Environment Id="UserPath" Action="set" Name="Path" Part="last" Permanent="no" System="no" Value="[INSTALLDIR]Toolchains\$(var.ProductVersion)+Asserts\usr\bin" />
+        <Environment Id="UserPath" Action="set" Name="Path" Part="last" Permanent="no" System="no" Value="[_usr_bin]" />
```
compnerd, line 331: Nice I was curious about this.  We likely want to have the same pattern everywhere.

compnerd (approved):

compnerd:
Kicked off a new CI run, 🤞 it passes!

compnerd:
The build failed: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/704/consoleText

Seems that the path to the clang headers is getting truncated @barnson?

barnson:
D'oh. Fixed.

compnerd:
New run kicked off: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/705/

barnson:
Embedded cabinets match what build-windows-toolchain.bat expects. We should revisit if we can prefer the bundle over the MSIs.

compnerd:
Re-running https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/706/

barnson:
💯

compnerd:
@barnson why is installer.exe 640MB?  That seems like a pretty hefty size increase from 399MB.

barnson:
Silent install doesn't require any gesture to agree with the EULA. Is it needed? I can default the checkbox to checked or get rid of it entirely when I do the custom theme.

compnerd:
> Silent install doesn't require any gesture to agree with the EULA. Is the checkbox needed? I can default the checkbox to checked or get rid of it entirely when I do the custom theme.

No, I don't think that the checkbox is needed.  We can deal with that aspect later :).

compnerd:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/708/ build succeeded, seems reasonable (should smoke test it).  Need to wait for the parallel change to be ready before we can merge this.

PR #237, opened 2023-08-30
WiX: adjust the packaging for CxxStdlib

compnerd:
We should use the version emitted into the `swift` rather than `swift_static` since we are linking against the dynamically linked variant


egorzhdan (approved):

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/798/

PR #236, opened 2023-08-30
WiX: adjust the installer UI

compnerd:
Adjust the L&F to maintain the aspect ratio that we previously had (4:3).  This also adjusts some of the spacing to make it more uniform and adjusts the location of the side banner.  The banner is currently shorter than the desired height, but is expected to be replaced with an appropriately sized image soon.  It also enables the feature selection items for the ARM64 and X86 SDKs.

PR #238, opened 2023-08-30
WiX: taller side artwork

lxbndr:
For upcoming L&F changes in #236


compnerd (approved):

lxbndr:
Updated with 4x size for hi-dpi displays

PR #241, opened 2023-08-31
WiX: package Observable into the distribution

compnerd:
Include the Observable module in the SDK and runtime.

### DIFF
--- a/platforms/Windows/rtl/lib/rtllib.wxs
+++ b/platforms/Windows/rtl/lib/rtllib.wxs
@@ -42,6 +42,9 @@
         <File Source="$(SDK_ROOT)\usr\bin\swiftCRT.dll" />
       </Component>
       <Component>
+        <File Source="$(SDK_ROOT)\usr\bin\swiftObservation.dll" />
+      </Component>
+      <Component>
         <File Source="$(SDK_ROOT)\usr\bin\swiftRemoteMirror.dll" />
       </Component>
       <Component>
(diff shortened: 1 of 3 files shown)


platforms/Windows/rtl/lib/rtllib.wxs
```diff
@@ -38,6 +38,9 @@
       <Component>
         <File Source="$(SDK_ROOT)\usr\bin\swiftCRT.dll" />
       </Component>
+      <Component>
+        <File Source="$(SDK_ROOT)\usr\bin\swiftObservation.dll" />
```
phausler, line 42: do we need to also provide the macros package too?

phausler (commented):
the library part seems reasonable location wise.

platforms/Windows/rtl/lib/rtllib.wxs
compnerd: I didn't know about the macros package!  Could you point me to that?  We will want to do that, but I think that for now it doesn't matter as macros aren't available yet on Windows (but working towards it).

phausler: the macro is created from here: https://github.com/apple/swift/tree/main/lib/Macros/Sources/ObservationMacros

phausler (approved):
it looks like all the appropriate parts are included; I am not familiar with the wxs format so I can't comment on the soundness of that part.

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/1018/

PR #250, opened 2023-10-13
Install the VC++ redistributable as part of the Swift toolchain installer

tristanlabelle:
Packages the VC++ redistributable and installs it as part of the toolchain install on Windows, since this is a dependency. Note that this will pop a UAC prompt because that installer is per-machine.

This is the [Microsoft-recommended approach](https://learn.microsoft.com/en-us/cpp/windows/deployment-in-visual-cpp?view=msvc-170): run the vcredist installer to install it globally (which makes it serviceable by microsoft), and don't uninstall it when uninstalling the toolchain since other apps may depend on it.

This requires [build script changes](https://github.com/compnerd/swift-build/pull/664) to take effect, but is backwards compatible with build scripts that don't specify the new properties.

I consciously departed from `ALL_CAPS` naming of properties because the MSBuild standard is `PascalCase` and the logic in `installer.wixproj` was looking really ugly otherwise.

# How tested:
Built with and without build script changes, uninstalled the VC runtime, installed the Swift toolchain, and verified that the VC runtime was installed in Add/Remove programs:

image
image

### DIFF
--- a/platforms/Windows/bundle/installer.wxs
+++ b/platforms/Windows/bundle/installer.wxs
@@ -59,23 +59,14 @@
     -->
 
     <Chain>
-      <?if $(VCREDIST_INSTALLER) != ""?>
-        <ExePackage
-          SourceFile="$(VCREDIST_INSTALLER)"
+      <?if $(VCRedistInstaller) != ""?>
+        <BundlePackage
+          SourceFile="$(VCRedistInstaller)"
           InstallCondition="OptionsInstallRtl"
           Permanent="yes"
           InstallArguments="/install /quiet /norestart"
-          DownloadUrl="$(BaseReleaseDownloadUrl)/{2}">
-          <!--
-            Install only if not already in Add/Remove Programs
-            under the GUID below.
-            TODO: Is this the right GUID for arm64?
-          -->
-          <ArpEntry 
-            Id="9e10fdb5-a50b-4c3b-98a4-aefac04e5970"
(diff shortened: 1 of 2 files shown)


platforms/Windows/bundle/installer.wxs
```diff
@@ -59,6 +59,25 @@
     -->
 
     <Chain>
+      <?if $(VCREDIST_INSTALLER) != ""?>
+        <ExePackage
```
barnson, line 63: I recommend using `BundlePackage` instead; it handles detection and versioning for you. Right now, the redist would be installed every time, which would lead to failure when somebody has a later version already installed.

tristanlabelle: Great to know, thanks a lot for the tip! I will investigate

tristanlabelle: I migrated to `<BundlePackage>` (and using aka.ms as the download url). Thanks!

compnerd:
All uppercase names are public properties, the pascal case variables are private.

compnerd (approved):

compnerd:
Tested at https://github.com/apple/swift/pull/69182 - https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/947/

compnerd:
We should backport this to 5.10

compnerd:
https://github.com/apple/swift-installer-scripts/pull/251

PR #252, opened 2023-10-25
WiX: avoid using extended versioning on the MSIs

tristanlabelle:
`<Package>` versions must be merely uint16's with no trailing `-`. We used to handle this in `build.ps1` on a per-wixproj basis before the refactoring that made `installer.wixproj` drive the whole build. The `<Bundle>` version still keeps using the semantic version suffix because that is supported.


compnerd (approved):
Thanks!

PR #255, opened 2023-10-31
WiX: no longer honour `OptionsInstallRtl`

compnerd:
The runtime library must be installed as the Swift compiler depends on this for the macros support as well as the driver being written in Swift. This is a hard dependency for the build tools.


etcwilde (approved):

tristanlabelle (approved):

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/983/

PR #257, opened 2023-11-03
WiX: add `SwiftIDEUtils.dll` to the image

compnerd:
This adds an additional DLL to the image.  The compiler is now able to vend enough of the swift-syntax library to enable building the LSP without adding a second copy of the parser.  This shaves ~6MiB off of the LSP binary.


etcwilde (approved):
LGTM. This is the SwiftIDEUtils from swift-syntax, yeah?

compnerd:
@etcwilde it is indeed; which is needed to swap out the static linking of swift-syntax and its second build.

PR #273, opened 2024-01-12
WiX: package and distribute `clang-scan-deps`

compnerd:
This is scanned for and used by newer CMake releases. Include it in the toolchain distribution packaging.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/1107/

hyp (approved):

PR #274, opened 2024-01-17
WiX: Remove _RegexParser's .swiftinterface from the SDK.

tshortli:
The module is becoming non-resilient in https://github.com/apple/swift/pull/70953.


compnerd (approved):
Do we need to worry about 5.10?

tshortli:
No, I was only planning to make this change on main.

PR #279, opened 2024-02-23
ICU: tweak build rules slightly

compnerd:
- Emit the stub data library into a subdirectory to prevent collisions
- Remove the generated output from byproducts, it is the output


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-windows/22082/

hyp (approved):
LGTM

PR #290, opened 2024-05-09
ICU: explicitly suffix tools for cross-compiling support

compnerd:
When cross-compiling to a different platform from Windows, we should explicitly suffix the executable binary for invocation.


hyp (approved):
Thanks!

PR #291, opened 2024-05-13
Revert "[icu] not building BUILD_DATA should still have an install ta…

hyp:
…rget"

This reverts commit 4c12ddfd870ad9bc80057e13a3169f42975f5d7b.

Revert "[icu] add a BUILD_DATA flag to let android build skip building data"

This reverts commit 955e5a9cfe5ff5bcb1cdf35cbd84583b73359c73.

This wasn't the right approach for ICU for android on Windows.


compnerd (commented):
I'm confused, shouldn't the `install` bits remain and the condition around it be removed? (along with the early return if `BUILD_DATA` is defined).

compnerd (approved):
Discussed offline - there is a second copy of the `install` call at the end of the file that accomplishes this.

PR #297, opened 2024-05-22
Package and bundle Android SDKs into the windows toolchain installer

hyp:
This change packages the Android SDKs that can be built using build.ps1 on windows into MSIs, which are then added to the toolchain's installer.

At the moment it covers aarch64 only, I'm adding and testing other arches and will update the PR with them as well.

For future follow-up:
- Fix the 'CxxStdlib' build issue for Android and add it to the installer. (https://github.com/apple/swift/issues/73855)

### DIFF
--- a/platforms/Windows/SideBySideUpgradeStrategy.props
+++ b/platforms/Windows/SideBySideUpgradeStrategy.props
@@ -26,6 +26,9 @@
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'amd64' ">{762D10FE-EBE5-4554-BB78-FB13A4A487E3}</SdkUpgradeCode>
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'arm64' ">{9749D9E6-E860-4FF6-9E8A-525270F471A3}</SdkUpgradeCode>
     <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'aarch64' ">{485f88f4-9342-48cb-853a-12da885a5818}</AndroidSdkUpgradeCode>
+    <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'x86_64' ">{0838ee60-5d4a-4832-b844-73dad6eb1cc1}</AndroidSdkUpgradeCode>
+    <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'armv7' ">{1269a926-3528-4ab7-b4d6-386d5c3f903a}</AndroidSdkUpgradeCode>
+    <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'i686' ">{d889349b-0000-4600-a04a-93602525d5db}</AndroidSdkUpgradeCode>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(MajorMinorProductVersion)' == '0.0'">
--- a/platforms/Windows/SideBySideUpgradeStrategy.props
+++ b/platforms/Windows/SideBySideUpgradeStrategy.props
@@ -25,6 +25,7 @@
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'x86' ">{443F4D7F-38F3-47C8-9BEE-37FEB01D13C8}</SdkUpgradeCode>
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'amd64' ">{762D10FE-EBE5-4554-BB78-FB13A4A487E3}</SdkUpgradeCode>
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'arm64' ">{9749D9E6-E860-4FF6-9E8A-525270F471A3}</SdkUpgradeCode>
+    <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'aarch64' ">{485f88f4-9342-48cb-853a-12da885a5818}</AndroidSdkUpgradeCode>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(MajorMinorProductVersion)' == '0.0'">
@@ -53,6 +54,7 @@
       IdeUpgradeCode=$(IdeUpgradeCode);
       RtlUpgradeCode=$(RtlUpgradeCode);
       SdkUpgradeCode=$(SdkUpgradeCode);
+      AndroidSdkUpgradeCode=$(AndroidSdkUpgradeCode);
     </DefineConstants>
   </PropertyGroup>
 </Project>
--- a/platforms/Windows/SideBySideUpgradeStrategy.props
+++ b/platforms/Windows/SideBySideUpgradeStrategy.props
@@ -22,13 +22,13 @@
     <DbgUpgradeCode>{91D382AF-1E92-44DC-A4AD-AEE91C1B5160}</DbgUpgradeCode>
     <IdeUpgradeCode>{8DD91C86-D13D-490B-B06B-9522A9CF504C}</IdeUpgradeCode>
     <RtlUpgradeCode>{BEA8C6DC-F73E-445B-9486-2333D1CF2886}</RtlUpgradeCode>
-    <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'x86' ">{443F4D7F-38F3-47C8-9BEE-37FEB01D13C8}</SdkUpgradeCode>
-    <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'amd64' ">{762D10FE-EBE5-4554-BB78-FB13A4A487E3}</SdkUpgradeCode>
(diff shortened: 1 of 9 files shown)


platforms/Windows/SideBySideUpgradeStrategy.props
```diff
@@ -25,6 +25,7 @@
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'x86' ">{443F4D7F-38F3-47C8-9BEE-37FEB01D13C8}</SdkUpgradeCode>
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'amd64' ">{762D10FE-EBE5-4554-BB78-FB13A4A487E3}</SdkUpgradeCode>
     <SdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'arm64' ">{9749D9E6-E860-4FF6-9E8A-525270F471A3}</SdkUpgradeCode>
```
compnerd, line 27: Should we rename these to `WindowsSDKUpgradeCode`?

platforms/Windows/SideBySideUpgradeStrategy.props
```diff
@@ -25,6 +25,7 @@
...
+    <AndroidSdkUpgradeCode Condition=" '$(ProductArchitecture)' == 'aarch64' ">{485f88f4-9342-48cb-853a-12da885a5818}</AndroidSdkUpgradeCode>
```
compnerd, line 28: And match this as `AndroidSDKUpgradeCode`. I think that we should add the other architectures as well.

platforms/Windows/bundle/installer.wixproj
```diff
@@ -31,4 +32,8 @@
   <ItemGroup Condition=" '$(INCLUDE_ARM64_SDK)' != '' ">
     <ProjectReference Include="..\sdk\sdk.wixproj" Properties="ProductArchitecture=arm64;Platform=x86" BindName="sdk_arm64" />
   </ItemGroup>
+
+  <ItemGroup Condition=" '$(INCLUDE_ANDROID_SDKS)' != '' ">
```
compnerd, line 36: Should this not be architecture specific?

platforms/Windows/bundle/installer.wxs
```diff
@@ -119,6 +120,15 @@
           <MsiProperty Name="INSTALLREDIST" Value="[OptionsInstallRedistArm64]" />
         </MsiPackage>
       <?endif?>
+
+      <?if $(INCLUDE_ANDROID_SDKS) == true ?>
```
compnerd, line 124: We should allow for this to extend to other architectures like we did for Windows.

platforms/Windows/android_sdk/android_sdk.wxs
```diff
@@ -0,0 +1,512 @@
...
+      <ComponentGroupRef Id="BlocksRuntime" />
+      <ComponentGroupRef Id="libdispatch" />
+      <ComponentGroupRef Id="_Concurrency" />
+      <ComponentGroupRef Id="_Differentiation" />
+      <ComponentGroupRef Id="_RegexParser" />
+      <ComponentGroupRef Id="_StringProcessing" />
+      <ComponentGroupRef Id="_Volatile" />
+      <ComponentGroupRef Id="Android" />
+      <ComponentGroupRef Id="Cxx" />
+      <!-- FIXME: <ComponentGroupRef Id="CxxStdlib" /> -->
```
compnerd, line 493: Is there a GH issue associated with this?

platforms/Windows/bundle/installer.wixproj
hyp: what's are the advantages of allowing to use arch-specific options/config variables instead of single switch for all android SDKs?

hyp: or should we have just one user-facing option, but let the installer configure which SDKs to include? I guess that would make sense

compnerd: Windows allows the user to select the architectures they want to target. It does allow for some space reduction for the installed toolchain. More importantly, it is setting up support for an online installer that would reduce the overall download size.

platforms/Windows/android_sdk/android_sdk.wxs
hyp: filing one

hyp: https://github.com/apple/swift/issues/73855

platforms/Windows/bundle/installer.wxs
```diff
@@ -38,6 +38,10 @@
     <Variable Name="OptionsInstallRedistAMD64" bal:Overridable="yes" Persisted="yes" Value="1" />
     <Variable Name="OptionsInstallSdkArm64" bal:Overridable="yes" Persisted="yes" Value="1" />
     <Variable Name="OptionsInstallRedistArm64" bal:Overridable="yes" Persisted="yes" Value="1" />
+    <Variable Name="OptionsInstallAndroidSdkArm64" bal:Overridable="yes" Persisted="yes" Value="1" />
```
tristanlabelle, line 41: Is the default value specified here? I think the Android SDK should be opt-in.

tristanlabelle (approved):
Looking great! I think we should make installing the Android SDK opt-in though. Most devs out there will likely never need it.

platforms/Windows/bundle/theme.xml
```diff
@@ -9,7 +9,7 @@
     <Font Id="2" Height="-22" Weight="500" Foreground="graytext">Segoe UI</Font>
     <Font Id="3" Height="-12" Weight="500" Foreground="windowtext" Background="window">Segoe UI</Font>
 
-    <Window Width="614" Height="384" HexStyle="100a0000" FontId="0" Caption="#(loc.Caption)">
+    <Window Width="614" Height="456" HexStyle="100a0000" FontId="0" Caption="#(loc.Caption)">
         <ImageControl X="0" Y="0" Width="165" Height="384" ImageFile="swift_side.png"/>
```
lxbndr, line 13: New side banner from ##298 would need this height to be adjusted as well
```suggestion
        <ImageControl X="0" Y="0" Width="165" Height="456" ImageFile="swift_side.png"/>
```

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -16,13 +16,21 @@
   <String Id="Sdk_ProductName_arm64" Value="Swift Windows SDK (ARM64)" />
   <String Id="Sdk_ProductName_amd64" Value="Swift Windows SDK (AMD64)" />
   <String Id="Sdk_ProductName_x86" Value="Swift Windows SDK (X86)" />
+  <String Id="AndroidSdk_ProductName_aarch64" Value="Swift Android SDK (Arm64)" />
```
tristanlabelle, line 19: The suffixes are inconsistent with elsewhere. Why not use the same? Is it because we're formatting the string elsewhere?

Also the casing of Arm64 is different.

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -16,13 +16,21 @@
...
+  <String Id="AndroidSdk_ProductName_armv7" Value="Swift Android SDK (ARM)" />
+  <String Id="AndroidSdk_ProductName_i686" Value="Swift Android SDK (X86)" />
   <String Id="BundleName" Value="Swift Developer Toolkit" />
   <String Id="ManufacturerName" Value="swift.org" />
 
   <String Id="OptionsFeaturesLabel" Value="Feature Selection" />
   <String Id="Redist_arm64" Value="Swift Windows Redistributable (ARM64)" />
   <String Id="Redist_amd64" Value="Swift Windows Redistributable (AMD64)" />
   <String Id="Redist_x86" Value="Swift Windows Redistributable (X86)" />
+  <String Id="Android_Sdk_arm64" Value="Swift Android SDK (ARM64)" />
```
tristanlabelle, line 30: Any reason to not reference the same string used above?

tristanlabelle (approved):
Looking good!

platforms/Windows/bundle/theme.xml
hyp: Tested it , works great!

hyp:
tested in https://github.com/apple/swift/pull/74220

hyp:
@compnerd could you take another look?

platforms/Windows/bundle/installer.wixproj
```diff
@@ -8,6 +8,10 @@
       $(DefineConstants);
       INCLUDE_X86_SDK=$(INCLUDE_X86_SDK);
       INCLUDE_ARM64_SDK=$(INCLUDE_ARM64_SDK);
+      INCLUDE_ARM64_ANDROID_SDK=$(INCLUDE_ARM64_ANDROID_SDK);
+      INCLUDE_x86_64_ANDROID_SDK=$(INCLUDE_x86_64_ANDROID_SDK);
+      INCLUDE_ARM_ANDROID_SDK=$(INCLUDE_ARM_ANDROID_SDK);
+      INCLUDE_X86_ANDROID_SDK=$(INCLUDE_X86_ANDROID_SDK);
```
compnerd, line 14: It would be nicer to swap this around to `ANDROID_...` IMO

compnerd (approved):
Minor tweak would be nice to have.

PR #304, opened 2024-06-06
[Windows] Add _CompilerSwiftSyntax libs and plugin servers

rintaro:
Update for https://github.com/apple/swift/pull/73725


compnerd (approved):
The changes look fine, but we should smoke test the build before we merge this.

PR #308, opened 2024-06-25
[icu] set CMAKE_POSITION_INDEPENDENT_CODE to YES

hyp:


compnerd (approved):
Yeah, this is required for non-Windows platforms.

compnerd:
I think that I want to run a quick test to ensure that this doesn't regress anything (I don't expect it to) before merging.

PR #317, opened 2024-07-30
[Windows/Android] Add swift-testing

rintaro:


compnerd (approved):
This seems fine from the packaging side, but I assume that there is a SPM change associated with this? Please do a `@swift-ci please build toolchain windows platform` test on swift with this PR.

PR #324, opened 2024-08-06
[windows] add _CompilerSwiftIfConfig dependency to the swift-syntax i…

hyp:
…nstaller step

https://github.com/swiftlang/swift/pull/75014 added the new dependency, that the swift compiler depends on, and thus we need to make sure that this DLL is installed


compnerd (approved):

DougGregor:
Thank you, @hyp!

PR #325, opened 2024-08-08
WiX: add Foundation macros to the build tools

compnerd:
Include a prebuilt copy of the Foundation macros in the toolchain distribution on Windows to match the desired behaviour.

### DIFF
--- a/platforms/Windows/bld/bld.wxs
+++ b/platforms/Windows/bld/bld.wxs
@@ -335,6 +335,12 @@
       </Component>
     </ComponentGroup>
 
+    <ComponentGroup Id="FoundationMacros" Directory="_usr_bin">
+      <Component>
+        <File Source="$(TOOLCHAIN_ROOT)\usr\bin\FoundationMacros.dll" />
+      </Component>
+    </ComponentGroup>
+
     <ComponentGroup Id="argument_parser" Directory="_usr_bin">
       <Component>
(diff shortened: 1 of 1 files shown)


compnerd:
Cross-tested in swiftlang/swift#75785 with https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5419/

platforms/Windows/bld/bld.wxs
```diff
@@ -335,6 +335,12 @@
       </Component>
     </ComponentGroup>
 
+    <ComponentGroup Id="FoundationMacros" Directory="_usr_bin">
+      <Component>
+        <File Source="$(TOOLCHAIN_ROOT)\usr\bin\FoundationMacros.dll" />
```
jmschonfeld, line 340: Is this the right path for the macro library? I'd expect it to be under a `host\plugins` folder somewhere, but if this is where the rest of the stdlib macros already live then that's fine I suppose assuming the compiler knows how to find this folder to add to `-plugin-path`

compnerd: The DLLs are in "bin", "lib" is where the import library lives. The plugin path makes sense, but I'm mirroring the path where the standard library macros are currently residing (e.g. observation macros). I think that I can do a follow up to move things around.

compnerd: Also note that #317 also put the macros into /usr/bin.

jmschonfeld (approved):

PR #329, opened 2024-08-15
WiX: correct the install conditions

compnerd:
This is a surprising interaction between variable types. Variables set at the command line are set as string types. The variables in the bundle are numeric (either `1` or `0` to match the semantics of check boxes). The original install condition
(`InstallCondition="OptionsInstall.*"` would evaluate the condition as a string, where _any_ value would evaluate as `true`. To force a comparison, where Burn will coerce values more like expectations, perform an explicit equality check.

### DIFF
--- a/platforms/Windows/bundle/installer.wxs
+++ b/platforms/Windows/bundle/installer.wxs
@@ -77,21 +77,21 @@
 
       <MsiPackage
         SourceFile="!(bindpath.cli)\cli.msi"
-        InstallCondition="OptionsInstallCli"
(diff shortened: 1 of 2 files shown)


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5422/ ; ran the installer locally and verified that `.\installer.exe /quiet OptionsInstallIde=0` did not install `sourcekit-lsp.exe`.

hyp (approved):

PR #332, opened 2024-08-19
WiX: extract Testing Macros into a component (NFC)

compnerd:
Extract the testing macros into their own component group. Each repository is currently setup into a single component group that we then enumerate into the install image. This follows suit for the testing macros.


compnerd:
CC: @rintaro @hyp

PR #336, opened 2024-09-05
[windows] add usr/lib/swift/clang to the toolchain installer

hyp:
this directory contains clang's resource headers that are needed by the clang embedded inside of the swift compiler, as it assumes its clang resources are in usr/lib/swift/clang, as opposed to standalone clang looking in usr/lib/clang/<clang_version>


hyp:
Testing it right now.

compnerd (approved):
This is rather unfortunate.

stevapple:
It’s a pity that we didn’t have it fixed [two years ago](https://github.com/swiftlang/swift-installer-scripts/pull/144)😳

compnerd:
@stevapple yeah, it still is not needed for Windows itself, but is needed for cross-compilation. I think that we need to find a better solution here still, but this is a stop gap to get the Android cross-compilation support working.

hyp:
I validated that this change installs the directory as expected in the toolchain MSI.

PR #360, opened 2025-01-03
[Windows] Add '_CompilerSwiftLexicalLookup.dll'

rintaro:
Update for https://github.com/swiftlang/swift/pull/78335


compnerd (approved):
This might need to be ported to 6.1 as well.

bnbarham:
> This might need to be ported to 6.1 as well.

It's not in 6.1 yet, but if we add it, then yeah 👍

bnbarham (approved):

PR #364, opened 2025-01-20
Install the SourceKit plugin into Windows toolchains

ahoppen:
Install the SourceKit plugin, added by https://github.com/swiftlang/sourcekit-lsp/pull/1906 into the Windows toolchains.

rdar://142909870

### DIFF
--- a/platforms/Windows/ide/ide.wxs
+++ b/platforms/Windows/ide/ide.wxs
@@ -37,6 +37,14 @@
         <File Source="$(TOOLCHAIN_ROOT)\usr\lib\sourcekitdInProc.lib" />
       </Component>
 
+      <Component Directory="_usr_bin">
+        <File Source="$(TOOLCHAIN_ROOT)\usr\lib\SwiftSourceKitPlugin.dll" />
+      </Component>
+
+      <Component Directory="_usr_bin">
+        <File Source="$(TOOLCHAIN_ROOT)\usr\lib\SwiftSourceKitClientPlugin.dll" />
+      </Component>
+
       <Component Directory="_usr_include_SourceKit">
         <File Source="$(TOOLCHAIN_ROOT)\usr\include\SourceKit\sourcekitd.h" />
       </Component>


platforms/Windows/ide/ide.wxs
```diff
@@ -37,6 +37,14 @@
         <File Source="$(TOOLCHAIN_ROOT)\usr\lib\sourcekitdInProc.lib" />
       </Component>
 
+      <Component Directory="_usr_lib">
+        <File Source="$(TOOLCHAIN_ROOT)\usr\lib\SwiftSourceKitClientPlugin.dll" />
+      </Component>
+
+      <Component Directory="_usr_lib">
+        <File Source="$(TOOLCHAIN_ROOT)\usr\lib\SwiftSourceKitClientPlugin.dll" />
+      </Component>
```
compnerd, line 46: Why is this duplicated?

platforms/Windows/ide/ide.wxs
```diff
@@ -37,6 +37,14 @@
         <File Source="$(TOOLCHAIN_ROOT)\usr\lib\sourcekitdInProc.lib" />
       </Component>
 
+      <Component Directory="_usr_lib">
```
compnerd, line 40: This seems like the wrong place ... DLLs are placed into the `bin` directory, `lib` should have static libraries and import libraries.

compnerd (changes requested):

platforms/Windows/ide/ide.wxs
ahoppen: You mean, why I’ve got two `Component` clauses pointing at `_usr_lib`? Looking at the other install locations, I couldn’t find any examples of two `File` entries in one component, so I thought that that’s the pattern.

Also note that this is SwiftSourceKit**Client**Plugin.dll and not SwiftSourceKitPlugin.dll

platforms/Windows/ide/ide.wxs
ahoppen: These libraries will be opened by SourceKit using `dlopen` and are not directly linked to SourceKit. Hence, they don’t ned to go into `bin` for Windows to pick them up and I thus thought that it makes more sense to put them in `lib`. Happy to change it, if you prefer to have them in `bin`.

platforms/Windows/ide/ide.wxs
compnerd: Sorry, maybe I'm misreading it, but the source file that is being installed is the _same_ file, so you have two entries for the same file.

platforms/Windows/ide/ide.wxs
compnerd: I think that they should be in `bin`. Unless they are completely standalone (including any system library dependencies). The current directory is scanned for any dependencies, and having them elsewhere makes that difficult.

platforms/Windows/ide/ide.wxs
ahoppen: Oh, whoops, they are indeed the same. Thanks for noticing the copy-paste error.

platforms/Windows/ide/ide.wxs
ahoppen: These libraries are fully standalone and don’t depend on any of the other libraries in `bin`.

compnerd: But that also includes the C/C++ runtimes. They _can_ also be redistributed in the toolchain (and if we run into the cases where we need to, we would fall back to that). In the mean time, note that other "plugin" libraries are loaded from `bin` on Windows.

ahoppen: Ah, I see. Moved it to `bin`

compnerd (approved):

compnerd:
Please do a cross-repository test with `@swift-ci please build toolchain Windows platform` before merging.

ahoppen:
Triggered a toolchain build: https://github.com/swiftlang/swift/pull/37710#issuecomment-2613377798

bnbarham:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5672/ 🟢

ahoppen:
Verified that the produced toolchain contains `SwiftSourceKitPlugin.dll` and `SwiftSourceKitClientPlugin.dll`.

PR #369, opened 2025-01-24
Default InstallerPlatform to x86 if unset

Steelskin:
If unset, `InstallerPlatform` defaults to `Platform`, which may be set to an invalid value for Android. This sets `InstallerPlatform` to `x86` if it is unset from the command line, allowing for custom overrides.

### DIFF
--- a/platforms/Windows/Directory.Build.props
+++ b/platforms/Windows/Directory.Build.props
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <ProductArchitecture Condition=" '$(ProductArchitecture)' == '' ">amd64</ProductArchitecture>
     <ProductVersion Condition=" '$(ProductVersion)' == '' ">0.0.0</ProductVersion>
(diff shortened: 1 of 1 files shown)


platforms/Windows/Directory.Build.props
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <ProductArchitecture Condition=" '$(ProductArchitecture)' == '' ">amd64</ProductArchitecture>
     <ProductVersion Condition=" '$(ProductVersion)' == '' ">0.0.0</ProductVersion>
+    <InstallerPlatform Condition=" '$(InstallerPlatform)' == '' ">x86</InstallerPlatform>
```
compnerd, line 6: I think that we should be setting this at the SDK level, not at the global level. This would otherwise install the ARM64 slice of the runtime library into the wrong architecture.

Steelskin: Following offline discussion, this should be set by the command line invocation. Closing.

PR #373, opened 2025-01-31
WiX: adjust manifest for static conversion

compnerd:
Update the SPM file list after some of the modules were converted to static linking.

### DIFF
--- a/platforms/Windows/cli/cli.wxs
+++ b/platforms/Windows/cli/cli.wxs
@@ -208,12 +208,6 @@
         <File Source="$(TOOLCHAIN_ROOT)\usr\bin\Commands.dll" />
       </Component>
       <Component>
-        <File Source="$(TOOLCHAIN_ROOT)\usr\bin\CoreCommands.dll" />
-      </Component>
-      <Component>
-        <File Source="$(TOOLCHAIN_ROOT)\usr\bin\DriverSupport.dll" />
-      </Component>
-      <Component>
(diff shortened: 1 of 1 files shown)


platforms/Windows/cli/cli.wxs
```diff
@@ -222,9 +216,6 @@
       <Component>
         <File Source="$(TOOLCHAIN_ROOT)\usr\bin\PackageModel.dll" />
       </Component>
-      <Component>
-        <File Source="$(TOOLCHAIN_ROOT)\usr\bin\PackageModelSyntax.dll" />
```
etcwilde, line 226: It looks like `PackageModelSyntax` is still able to go either way in https://github.com/swiftlang/swift-package-manager/pull/8268/files and https://github.com/swiftlang/swift-package-manager/blob/2215d098ed2aa991ed3f3d8c712266f14ebe4b45/Sources/PackageModelSyntax/CMakeLists.txt#L9

Does that PR need to switch it to a static archive, or are you switching things to use `-DBUILD_SHARED_LIBS=NO`?

compnerd: Nice catch! That should still be dynamic - I believe it ended up being that this was not profitable.

etcwilde: Okay, now it does.

etcwilde (approved):

PR #388, opened 2025-02-22
WiX: extract CDispatch module definition

compnerd:
The core dispatch headers are uniform across the SDKs. Ideally we would also be able to share the "os" headers, but they are installed into a separate directory and we cannot simply include them for the restructuring.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5730/

shahmishal (approved):

PR #390, opened 2025-02-22
WiX: add educational notes to the windows packaging

compnerd:
This packages up the educational notes for distribution with the toolchain. It is yet unclear if this should be broken out into an optioanl MSI to allow installation of the documentation on demand rather than as an always available option. This should at least ensure that we distribute the user facing documentation that can be accessed from the compiler.


shahmishal (approved):

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5733/

PR #391, opened 2025-02-22
WiX: add educational notes to the windows packaging

compnerd:
This packages up the educational notes for distribution with the toolchain. It is yet unclear if this should be broken out into an optioanl MSI to allow installation of the documentation on demand rather than as an always available option. This should at least ensure that we distribute the user facing documentation that can be accessed from the compiler.

### DIFF
--- a/platforms/Windows/bld/bld.wxs
+++ b/platforms/Windows/bld/bld.wxs
@@ -32,6 +32,12 @@
     <DirectoryRef Id="_usr_share">
       <Directory Id="_usr_share_clang" Name="clang" />
       <Directory Id="_usr_share_swift" Name="swift" />
+      <Directory Id="_usr_share_doc" Name="doc">
+        <Directory Id="_usr_share_doc_swift" Name="swift">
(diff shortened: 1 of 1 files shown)


shahmishal (approved):

compnerd:
Installer built at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5732/

DougGregor:
Is there no way to do globbing on `*.md` within that directory?

compnerd:
I find that the explicit listing is better, and more importantly, the globbing really does cause a lot of upgrade issues. We currently only glob in two places, and both of them have been where we have always seen issues with users. The problem is that globbing interferes with the natural deterministic hashing and so we really should prefer to explicitly list everything if possible.

DougGregor:
Hmm. My actual goal here is to avoid having to do paired commits to the swift and swift-installer-scripts repositories for each educational note. Can we somehow ingest a file from the swift repository?

compnerd:
That would greatly increase build times for us due to the size of the swift repository. Can we ingest a file from the installer scripts in the swift repository instead?

PR #392, opened 2025-02-25
[signing] Add support for signing with Azure

Steelskin:
Modern Windows signing requirements require using an Azure secure vault to sign packages, using `AzureSignTool`.
This change adds support for that tool, by using it if it is provided via the `AZURE_SIGN_TOOL` property. The "old" way of signing with a provided self-signed certificate via the `CERTIFICATE` and `PASSPHRASE` properties is still available.


Steelskin:
@compnerd PTAL

Steelskin:
Closing following offline discussion. Trusted Signing is the new new preferred way.

PR #397, opened 2025-03-04
[WiX] Remove an outdated educational note from Windows packaging.

hborla:
I'm removing this note in https://github.com/swiftlang/swift/pull/79509/commits/f1492fe55372712bf90c4b822b2705fddcaec5d2; the restriction was lifted in SE-0362.

Starting in Swift 6.2, educational note identifiers should be stable and should not be removed in later toolchains. However, Swift 6.2 will be the first toolchain version to make that promise, so it's okay to clean up this educational note now.

### DIFF
--- a/platforms/Windows/bld/bld.wxs
+++ b/platforms/Windows/bld/bld.wxs
@@ -274,9 +274,6 @@
 
     <ComponentGroup Id="SwiftEducationalNotes" Directory="_usr_share_doc_swift_diagnostics">
       <Component>
-        <File Source="$(TOOLCHAIN_ROOT)\usr\share\doc\swift\diagnostics\complex-closure-inference.md" />
-      </Component>
-      <Component>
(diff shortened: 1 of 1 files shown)


compnerd (approved):
I suppose that this is fine, though, I think that we could leave it in on main and only remove it on the release/6.1 branch as main is 6.2-dev.

We need to make sure this makes it to release/6.1 though.

hborla:
> I suppose that this is fine, though, I think that we could leave it in on main and only remove it on the release/6.1 branch as main is 6.2-dev.

I'm confused. I'm only removing the educational note on `main`, not on `release/6.1`.

compnerd:
Oh, sorry, I misunderstood the commit message. The first toolchain to guarantee the stability of edunotes is 6.2, not this feature. Ignore the message, LGTM!

etcwilde (approved):

PR #399, opened 2025-03-05
WiX: package and distribute `swift-synthesize-interface`

compnerd:
This adds a new utility to the toolchain - `swift-synthesize-interface`.
For now, consider it a debugging utility as it is not used during the
normal compilation pipeline and thus is not a build tool. This is not
something that is normally used during CI pipelines and does not fit
into the CLI tools. As this is not currently integrated into the IDE, we
don't consider it an IDE tool either.

### DIFF
--- a/platforms/Windows/dbg/dbg.wxs
+++ b/platforms/Windows/dbg/dbg.wxs
@@ -105,6 +105,12 @@
       </Component>
     </ComponentGroup>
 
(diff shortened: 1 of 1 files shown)


compnerd:
Tested at: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5762/artifact/build/artifacts/

shahmishal (approved):

PR #401, opened 2025-03-16
WiX: adjust manifest for thick module for Dispatch

compnerd:
Adjust the manifest rules to account for the fact that libdispatch now properly stages the swiftmodule in a thick module format. This avoids us having to shuffle files around when building the installer.


compnerd:
Associated with https://github.com/swiftlang/swift-corelibs-libdispatch/pull/860.

compnerd:
Closing this as we now expect the thick module and perform the shuffling outside of the manifest. As the pieces for the libdispatch build land, we should be able to remove that shuffling.

PR #403, opened 2025-03-19
WiX: package up the cross-module overlay

compnerd:
This adds the swift-testing cross-module overlay to the distribution on Windows.

### DIFF
--- a/platforms/Windows/sdk/drd/sdk.wxs
+++ b/platforms/Windows/sdk/drd/sdk.wxs
@@ -83,7 +83,9 @@
                 <Directory Name="lib">
                   <Directory Name="swift">
                     <Directory Name="android">
+                      <Directory Id="_Testing_Foundation.swiftmodule" Name="_Testing_Foundation.swiftmodule" />
                       <Directory Id="Testing_usr_lib_swift_android_ARCH" Name="$(Architecture)" />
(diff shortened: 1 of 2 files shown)


stmontgomery (approved):
Yep, this looks like what you'd need to get the cross-import overlay working. LGTM, thanks!

etcwilde (approved):

grynspan:
@compnerd Does this need to be cherry-picked?

etcwilde:
Approved with a note, I don't actually see an accepted proposal for cross-import libraries. The [Swift-Testing proposal](https://github.com/swiftlang/swift-evolution/blob/1d3006316d2d6c65a7c69faec1886f70f9cb7ba3/proposals/testing/0009-attachments.md?plain=1#L37) only references the pitch thread.

compnerd:
@grynspan yes, we should cherry-pick this.

grynspan:
@etcwilde Cross-import overlays haven't been approved as a supported feature; our use here is necessary because Swift Testing is part of the toolchain and there is a circular dependency between Foundation and Swift Testing if Swift Testing puts these interfaces directly in the main `Testing` module. @beccadax and colleagues approved this limited use of the feature because of the constraints we're under.

(I'm on leave and won't respond further right now—please ping Becca, @stmontgomery, or @briancroom if you have further questions.)

PR #405, opened 2025-03-25
WiX: simplify the Android ARMv7 rules

compnerd:
The Android ARMv7 build will properly install the modules at this point without a separate triple for the modules. Update the rules accordingly bringing the ARMv7 inline with the other Android SDKs. The only outlier remaining would be `ds2` which should adopt the same triple.

### DIFF
--- a/platforms/Windows/sdk/drd/sdk.wxs
+++ b/platforms/Windows/sdk/drd/sdk.wxs
@@ -3,11 +3,9 @@
 
   <?if $(ProductArchitecture) = "arm64"?>
     <?define Architecture = "aarch64"?>
-    <?define SwiftModuleTriple = "aarch64-unknown-linux-android"?>
     <?define Triple = "aarch64-unknown-linux-android"?>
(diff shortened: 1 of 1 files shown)


etcwilde (approved):
Hmm, this confused me a bit. I see that the `Triple` is used to store the module triple, and `SwiftModuleTriple` is used to store... something else?...

```
<File Source="$(PLATFORM_ROOT)\Developer\Library\Testing-$(ProductVersion)\usr\lib\swift\android\Testing.swiftmodule\$(Triple).swiftinterface" />
```

Would it make sense to rename all of the `Triple`s `SwiftModulTriple` instead so that it's clear that `Triple` is the `SwiftModuleTriple` instead of the target triple?

compnerd:
> Would it make sense to rename all of the Triples SwiftModulTriple instead so that it's clear that Triple is the SwiftModuleTriple instead of the target triple?

Not really worth it IMO. The variable will be entirely removed in the next round of installer changes.

PR #406, opened 2025-03-29
WiX: restructure the Windows installer

compnerd:
This restructures the Windows installer components to the following:
- bld
- cli
- dbg
- ide
- rtl

The SDKs for the various platforms have now been removed and replaced with the new platforms packages. There are currently two of them:
- android
- windows

Each one of these platform support packages support a set of architectures relevant to the OS. The platform MSI installs the shared view of the platform support, and each architecture slice adds the support for the specified architecture.

The parameters to the installer build have now been replaced with the single root path to the toolchain image and a set of paths for the Windows runtime installations to package. This is critical to resolving the previous issue with the unified installer build: we would clobber the paths as there was no android and windows platform differentiation.

The changes removed the need for the variants of the shared library as all the architectures are now included into a single MSI, simply split up amongst the CABs.


compnerd:
Restructured installer options dialog:
image

We now have 2 proper platforms, each with architecture specific slices. The core platform support only installs shared files.

shahmishal (approved):

jakepetroules (approved):

PR #410, opened 2025-04-09
WiX: remove workaround for mimalloc

compnerd:
Remove the extra library packaging which was a workaround for mimalloc adding a reference to foreign architecture modules.

### DIFF
--- a/platforms/Windows/bld/bld.wixproj
+++ b/platforms/Windows/bld/bld.wixproj
@@ -4,7 +4,6 @@
       $(DefineConstants);
       _USR_LIB_CLANG=$(ImageRoot)\Toolchains\$(ProductVersion)+Asserts\usr\lib\clang;
       _USR_LIB_SWIFT_CLANG=$(ImageRoot)\Toolchains\$(ProductVersion)+Asserts\usr\lib\swift\clang;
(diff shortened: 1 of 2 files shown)


compnerd:
Associated change: https://github.com/swiftlang/swift/pull/80681

Tested the ARM64 build on a Parallels VM

etcwilde (commented):
https://github.com/microsoft/mimalloc/issues/997 is still open.
Though it's in v3.0.3, that version appears to be a beta release.
> Latest release : v3.0.3 (beta) (2025-03-28).

Does this affect 6.1? If so, I'm a bit concerned about the implications of a fully released toolchain and SDK switching to using beta libraries in its runtimes.

Does this affect 6.2? If so, will the fix be in a proper release of mimalloc by the time 6.2 is released?

compnerd:
This isn't going to 6.1, its too late for that. Maybe 6.1.1, but I don't think that this is urgent there.

I'd like this to go to 6.2 as well. This library is only used by the _compiler_ not the runtime. The runtime still uses the system allocator. While it is interesting to consider the swap there, I don't think that is an immediate priority.

We are currently on 3.0.1, which was arguably "alpha". The big changes here are related to enabling ARM64 support.

etcwilde (approved):

PR #415, opened 2025-04-21
Android: Remove `_Builtins_float`

Steelskin:
`_Builtins_float` was added in #406 for the Android package. However, this module does not exist so this removes it.


compnerd (changes requested):
No, this was intentionally added - this should be built.

Steelskin:
Not the right fix, closing.

PR #418, opened 2025-05-06
Install the SwiftPM/Swift Build resources in /usr/share/pm

owenv:
Builds on https://github.com/swiftlang/swift-installer-scripts/pull/417 - only the second commit is new

Ensure the Windows installer includes the resources bundled in https://github.com/swiftlang/swift-build/pull/488

### DIFF
--- a/platforms/Windows/cli/cli.wxs
+++ b/platforms/Windows/cli/cli.wxs
@@ -28,6 +28,19 @@
       </Directory>
     </DirectoryRef>
 
+    <DirectoryRef Id="_usr_share">
+      <Directory Id="_usr_share_pm" Name="pm">
+        <Directory Id="_usr_share_pm_SWBAndroidPlatform" Name="SwiftBuild_SWBAndroidPlatform.resources" />
+        <Directory Id="_usr_share_pm_SWBApplePlatform" Name="SwiftBuild_SWBApplePlatform.resources" />
+        <Directory Id="_usr_share_pm_SWBCore" Name="SwiftBuild_SWBCore.resources" />
+        <Directory Id="_usr_share_pm_SWBGenericUnixPlatform" Name="SwiftBuild_SWBGenericUnixPlatform.resources" />
+        <Directory Id="_usr_share_pm_SWBQNXPlatform" Name="SwiftBuild_SWBQNXPlatform.resources" />
+        <Directory Id="_usr_share_pm_SWBUniversalPlatform" Name="SwiftBuild_SWBUniversalPlatform.resources" />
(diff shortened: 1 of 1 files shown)


compnerd (changes requested):
I'd rather that we explicitly list the files if possible. Do you expect the list to constantly change dramatically? The only place that we currently harvest is in the clang resources, which has continued to be a source of installation failures. I think that we should prefer to just enumerate the content that we are distributing.

owenv:
I don't expect the list to change dramatically, but it is added to fairly often, and it's unfortunate if we have to stage cross-repo changes every time that happens

owenv:
I updated this patch to explicitly list everything and am autogenerating the content via scripts in the Swift Build repo

platforms/Windows/cli/cli.wxs
```diff
@@ -270,6 +719,15 @@
...
       <ComponentGroupRef Id="llbuild" />
+      <ComponentGroupRef Id="SwiftBuild" />
+      <ComponentGroupRef Id="SWBAndroidPlatformResources" />
+      <ComponentGroupRef Id="SWBApplePlatformResources" />
+      <ComponentGroupRef Id="SWBCoreResources" />
+      <ComponentGroupRef Id="SWBGenericUnixPlatformResources" />
+      <ComponentGroupRef Id="SWBQNXPlatformResources" />
+      <ComponentGroupRef Id="SWBUniversalPlatformResources" />
+      <ComponentGroupRef Id="SWBWebAssemblyPlatformResources" />
+      <ComponentGroupRef Id="SWBWindowsPlatformResources" />
```
compnerd, line 730: I think that SwiftBuild should be its own ComponentGroup. Is there a plan to merge SwiftBuild into SPM (not as a dependency but repository-wise)?

platforms/Windows/shared/shared.wxs
```diff
@@ -52,6 +52,16 @@
                 <Directory Id="_usr_share_docc" Name="docc">
                   <Directory Id="_usr_share_docc_render" Name="render" />
                 </Directory>
+                <Directory Id="_usr_share_pm" Name="pm">
```
compnerd, line 55: Why is this in shared? I think that this should be in cli.wxs

compnerd (changes requested):

platforms/Windows/shared/shared.wxs
owenv: No particular reason, fixed it

owenv:
Built a working toolchain with the latest changes in https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5888/. Installed content all looks correct.

PR #420, opened 2025-05-11
WiX: package SwiftFixIt.dll on Windows

compnerd:
This is part of SPM, introduced in
swiftlang/swift-package-manager#8585. The newly minted dependency causes `swift build` to fail as it is missing the library in the installation (caught by CI at
https://github.com/thebrowsercompany/swift-build/actions/runs/14890246041).

### DIFF
--- a/platforms/Windows/cli/cli.wxs
+++ b/platforms/Windows/cli/cli.wxs
@@ -238,6 +238,9 @@
         <File Source="$(ToolchainRoot)\usr\bin\SPMBuildCore.dll" />
       </Component>
       <Component>
+        <File Source="$(ToolchainRoot)\usr\bin\SwiftFixIt.dll" />
+      </Component>
+      <Component>
         <File Source="$(ToolchainRoot)\usr\bin\Workspace.dll" />
       </Component>
 


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5884/

etcwilde (approved):

xedin:
@compnerd I discussed with this @etcwilde and we should rather fix this on the swiftpm side and make SwiftFixIt static.

compnerd:
@xedin how long would that take? Windows hasn't had a toolchain bump since 4/3, and I'd rather get this in to unblock that.

xedin:
@compnerd a couple of hours for CI to run but if that's pressing go ahead an merge, we can always revert later.

compnerd:
I'll wait until later today - if we get the conversion in, that is good. Merging this and that would then break the build as this will fail to build the installer once that library disappears

xedin:
I'll keep an eye on the status of that PR and ping you.

compnerd:
Merged the SPM change, so going to close this for now.

PR #422, opened 2025-05-14
WiX: adjust the disk ids for the windows sdk

compnerd:
Ensure that these files are packaged into the appropriate cabs. When supporting the online installer, selecting (or de-selecting) an architecture should allow us to shave off the download for the components that are architecture specific. Make the default disk id (1) explicit as we prepare to add an experimental SDK.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5899/

PR #423, opened 2025-05-14
Split options dialog into 4 steps

mhegazy:
We are getting strapped for space in the current installer options window. To give us more room, this change splits the Options Page into 4, each with a specific theme. 1. Install Root, 2. Tools, 3. Windows SDK components, and 4. Android SDK components.

See this [video](https://github.com/user-attachments/assets/94829f3f-6069-4917-94b7-45970bfee330) for the new flow after the change:

Changes include:
- Split into 4 pages
- Associated strings and layout changes
- Change label from "Options" to "Customize"
- Added "Back" button to allow for a smother navigation experience

### DIFF
--- a/platforms/Windows/shared/swift.en-us.wxl
+++ b/platforms/Windows/shared/swift.en-us.wxl
@@ -28,7 +28,6 @@
   <String Id="BundleName" Value="Swift Developer Toolkit" />
   <String Id="ManufacturerName" Value="swift.org" />
 
-  <String Id="OptionsFeaturesLabel" Value="Feature Selection" />
   <String Id="Redist_arm64" Value="Swift Windows Redistributable (ARM64)" />
   <String Id="Redist_amd64" Value="Swift Windows Redistributable (AMD64)" />
   <String Id="Redist_x86" Value="Swift Windows Redistributable (X86)" />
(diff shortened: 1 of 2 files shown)


mhegazy:
CC @compnerd

compnerd:
Hmm, can we make it a hyperlinked pagination? Something like:

Core
 - [ ] Build Tools
 - [ ] CLI Tools
 - [ ] Debugging Tools
 - [ ] ...

SDKs
 - Windows [link]
 - Android [link]

mhegazy:
> Hmm, can we make it a hyperlinked pagination? Something like:

We could. There is a "CommandLink" element, that is basically a hyper link that acts as a button. but i do not think it adds much value to clarity. this flow is not very different from what [git for windows](https://git-scm.com/downloads/win) does today. a series of questions, and then install page. I think a simple and familiar interface is better than trying to over engineer it

mhegazy:
for context @compnerd, it looks something like this: 
image 

which i do not find it very ascetically appealing

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -48,13 +47,21 @@
   <String Id="HelpCloseButton" Value="&amp;Close" />
   <String Id="InstallLicenseLinkText" Value="[WixBundleName] &lt;a href=&quot;#&quot;&gt;license terms&lt;/a&gt;." />
   <String Id="InstallAcceptCheckbox" Value="I &amp;agree to the license terms and conditions" />
-  <String Id="InstallOptionsButton" Value="&amp;Options" />
+  <String Id="InstallOptionsButton" Value="Customi&amp;ze" />
```
etcwilde, line 50: Can you remind me what the `&amp;` does here. It looks weird putting it in the middle of the word and just want to make sure that's intentional.

etcwilde (approved):
Overall LGTM.

Not sure how you want to do this, but it might be good to allow the user to select which platforms they want to install instead of pushing them through that page every time. If we ever added a Linux SDK on Windows, that would be another page on the installer so it might get cumbersome. Otherwise this looks pretty good.

compnerd:
> We could. There is a "CommandLink" element, that is basically a hyper link that acts as a button. but i do not think it adds much value to clarity. this flow is not very different from what [git for windows](https://git-scm.com/downloads/win) does today. a series of questions, and then install page. I think a simple and familiar interface is better than trying to over engineer it

The difference is that this is not like the git flow. This is explicitly an options dialog that the user needs to navigate to in order to select features. It is not about the defaults. The Git installer has a feature tree.

platforms/Windows/shared/swift.en-us.wxl
mhegazy: the shortcut character for the button. since there is a cancel on this form, "C" was taken, so pickled "z" instead.

mhegazy:
>The difference is that this is not like the git flow. This is explicitly an options dialog that the user needs to navigate to in order to select features.

we can make it so. we just need to make the first option page the first page in the installer instead of "Install".

 > It is not about the defaults. The Git installer has a feature tree.
 
so this is where it gets complicated. Git is just an msi (a single Package in Wix terms). so the built in Wix UI gives the feature tree for free. we are using a chain of packages (a Bundle), and for these there a handful of UI options we can choose from, and none is a feature tree. 

if we want a feature tree there are two options. one option is to use one of the MSIs and mark it with `DisplayInternalUI` to show the UI within the chainer UI. it gets dicey quickly since now we have to use this feature tree to represent other components. 
The other option is to build our own custom chainer. there are a few projects out there that can help with that (e.g. [wixsharp](https://github.com/oleg-shilo/wixsharp/blob/master/Documentation/wiki_images/FeatureItemRendering.png))

barnson:
An alternative if you want a feature tree is to use a [BAFunctions](https://www.nuget.org/packages/WixToolset.WixStandardBootstrapperApplicationFunctionApi) DLL to fill the tree control and respond to the notifications to set the properties that control package state.

It would be a shame to introduce a .NET dependency just for the installer.

PR #425, opened 2025-05-20
[Packaging] Add No-asserts toolchain to setup

mhegazy:
This is a follow up to https://github.com/swiftlang/swift/pull/81644 which adds build support for no-asserts toolchain variant along with the default asserts variant.

**Note to reviewers**: Reviewing this change one commit at a time should give you a better experience than looking at all changes at once.

**Changes**:
- Split `bld.msi`, `ide.msi`, `cli.msi` and `dbg.msi` into two separate msi's each. the current authoring with no changes (Same upgrade codes, and component names) goes to a `<product>.asserts.msi` and `<product>.asserts.cab`, and a new msi is created using the same authoring but new upgrade codes to `<product>.noasserts.msi` and `<product>.noasserts.cab`
- The new msi's are **only** built if the command line option is passed to build.cmd/build.ps1. this ensures that build time do not regress as a result of this change
- To enable that without duplication, the authoring is parametrized.; the authoring for  `bld.msi`, `ide.msi`, `cli.msi` and `dbg.msi` moves from `<product>.wxs` to `<product>.wxi` and is shared between the two msi variants. 
  - To see the differences few differences between the two msi's look at `asserts/<product>.asserts.wxs` vs `noasserts/<product>.noasserts.wxs`
 - Create a new folder structure for no-asserts toolchain, by duplicating all our folders under `ToolchainsVersioned` defined in `shared.wxs`
   - Since the no-assert toolchain build output is not a complete layout, i.e. not all needed files are built twice, only ones that are different between the two toolchain variants are built. we need to pull filles from both output folders to make a complete layout. 
 - Add a new UI option to select one or both of the two variants of the toolchain
 - UI makes it so at least one toolchain should be enabled at anytime. users can choose to add both
    - Path env var is set to Asserts if it is selected. if users select **only** no-asserts variant, then the PATH will be set to this one.

**Still Open**
- [ ] UI for the installer. This changes extends the UI by 40px to add new options; this should be fixed in a separate PR before this change lands


platforms/Windows/SideBySideUpgradeStrategy.props
```diff
@@ -18,9 +18,13 @@
 
   <PropertyGroup>
     <BldUpgradeCode>{7E95DC06-7F84-4E8E-A038-8304AF0468FB}</BldUpgradeCode>
```
compnerd, line 20: We should rename these to indicate the `Asserts` variant.

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -25,6 +29,8 @@
   <String Id="Utl_ProductName_arm64" Value="Swift Windows Utilities (ARM64)" />
   <String Id="Utl_ProductName_amd64" Value="Swift Windows Utilities (AMD64)" />
   <String Id="Utl_ProductName_x86" Value="Swift Windows Utilities (X86)" />
+  <String Id="Asserts_Toolchain_ProductName" Value="Swift Toolchain (Asserts enabled)" />
+  <String Id="NoAsserts_Toolchain_ProductName" Value="Swift Toolchain (Asserts disabled)" />
```
compnerd, line 33: I think that we should just name the variant spelling here ...

```suggestion
  <String Id="Asserts_Toolchain_ProductName" Value="Swift Toolchain (Asserts)" />
  <String Id="NoAsserts_Toolchain_ProductName" Value="Swift Toolchain (NoAsserts)" />
```

I'm thinking about the next variant that would be useful to add: `Asserts+UBSan`, where `Asserts enabled, Undefined Behaviour Sanitizer enabled` seems like it would be pretty long.

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -3,9 +3,13 @@
 <WixLocalization xmlns="http://wixtoolset.org/schemas/v4/wxl" Culture="en-US">
   <String Id="DowngradeError" Value="A newer version of [ProductName] is already installed." />
   <String Id="Bld_ProductName" Value="Swift Build Tools" />
+  <String Id="BldNoAsserts_ProductName" Value="Swift Build Tools (Asserts disabled)" />
   <String Id="Cli_ProductName" Value="Swift Command Line Tools" />
+  <String Id="CliNoAsserts_ProductName" Value="Swift Command Line Tools (Asserts disabled)" />
   <String Id="Dbg_ProductName" Value="Swift Debugging Tools" />
+  <String Id="DbgNoAsserts_ProductName" Value="Swift Debugging Tools (Asserts disabled)" />
   <String Id="Ide_ProductName" Value="Swift IDE Integration Tools" />
+  <String Id="IdeNoAsserts_ProductName" Value="Swift IDE Integration Tools (Asserts disabled)" />
```
compnerd, line 12: Similarly, we should use the variant spelling and update the default ones to indicate the variant as well.

platforms/Windows/shared/shared.wxs
```diff
@@ -103,6 +130,10 @@
     <Property Id="TOOLCHAINSVERSIONED">
       <RegistrySearch Root="HKCU" Key="SOFTWARE\!(loc.ManufacturerName)\Installer\$(ProductVersion)\!(bind.Property.UpgradeCode)" Name="ToolchainsVersioned" Type="directory" />
```
compnerd, line 131: We should name this to include the variant.

platforms/Windows/shared/shared.wxs
```diff
@@ -86,6 +108,11 @@
         <RegistryValue Root="HKCU" Key="SOFTWARE\!(loc.ManufacturerName)\Installer\$(ProductVersion)\!(bind.Property.UpgradeCode)" Name="ToolchainsVersioned" Value="[ToolchainsVersioned]" />
         <util:RemoveFolderEx Property="TOOLCHAINSVERSIONED" Condition="TOOLCHAINSVERSIONED" />
       </Component>
+
+      <Component Id="VersionedDirectoryCleanupToolchains_noasserts">
```
compnerd, line 112: Can this not be merged into the singular `VersionedDirectoryCleanupToolchains`? We should be cleaning up `%ROOT%\Toolchains`.

platforms/Windows/shared/shared.wxs
```diff
@@ -59,6 +59,28 @@
...
+          </Directory>
+          <Directory Id="_usr_share_noasserts" Name="share">
+            <Directory Id="_usr_share_docc_noasserts" Name="docc">
+              <Directory Id="_usr_share_docc_render_noasserts" Name="render" />
+            </Directory>
+          </Directory>
+        </Directory>
+      </Directory>
+    </DirectoryRef>
+  </Fragment>
```
compnerd, line 82: I'm not sure I understand why the `_noasserts` suffix is needed at all. The layout is going to be identical, its just that the content source is different. The only piece that changes is the ToolchainsVersioned directory, which can include the suffix. If we do this, we can actually share the definition across the variants no?

platforms/Windows/shared/shared.wxs
mhegazy: that would break the upgrade scenario.

platforms/Windows/shared/shared.wxs
mhegazy: they are diffrent layouts. they are rooted at diffrent directores. one at `0.0.0+Asserts` and one at `0.0.0+NoAsserts`. a directory entry in an msi is just a name and parent. since the parents are diffrent, we need a new entry

platforms/Windows/SideBySideUpgradeStrategy.props
mhegazy: changed

platforms/Windows/shared/swift.en-us.wxl
mhegazy: changed

platforms/Windows/shared/swift.en-us.wxl
mhegazy: changed

platforms/Windows/shared/shared.wxs
mhegazy: no. they are two different directories:  `<INSTALLROOT>\Toolchains\0.0.0+Asserts` and `<INSTALLROOT>\Toolchains\0.0.0+NoAsserts`

platforms/Windows/bundle/installer.wixproj
```diff
@@ -8,7 +8,7 @@
       $(DefineConstants);
       IncludeAndroid=$(Platforms.Contains("android"));
       IncludeWindows=$(Platforms.Contains("windows"));
-      INCLUDE_NOASSERTS=$(INCLUDE_NOASSERTS);
+      Include_NoAsserts=$(ToolchainVariants.Contains("noasserts"));
```
compnerd, line 11: We slowly migrated from the `CAPS_UNDERSCORE` to the `TitleCase` variants, where don't have `_`.

platforms/Windows/shared/shared.wxs
compnerd: I see; lets make that a prefix I think.

platforms/Windows/shared/shared.wxs
compnerd: My understanding was that it was meant to clean up any orphaned directories in `<INSTALLROOT>\Toolchains`. Is that not the case?

mhegazy: there are two folders. Toolchain, and ToolchainVersioned:

```   
    <DirectoryRef Id="INSTALLROOT">
        <Directory Id="Toolchains" Name="Toolchains">
          <Directory Id="ToolchainsVersioned" Name="$(ProductVersion)+Asserts">
```
So for every top level folder we create we need to remember to remove it. so this is why we have similar entries for `VersionedDirectoryCleanupRedistVersion`, `VersionedDirectoryCleanupRuntimes` and `VersionedDirectoryCleanupRedistVersion`

platforms/Windows/shared/shared.wxs
mhegazy: why? all the other names it is a suffix. e.g. `dbg.noasserts`. why would component names and folder names be different?

platforms/Windows/bundle/installer.wixproj
mhegazy: fixed.

platforms/Windows/bld/noasserts/bld.noasserts..wxs
```diff
@@ -0,0 +1,11 @@
+﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
```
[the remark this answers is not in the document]
mhegazy: fixed.

platforms/Windows/bundle/installer.wxs
```diff
@@ -54,7 +54,16 @@
...
 
-    <!--
+    <Variable Name="OptionsInstallAssertsToolchain" bal:Overridable="yes" Persisted="yes" Value="1" />
+    <Variable Name="OptionsInstallNoAssertsToolchain" bal:Overridable="yes" Persisted="yes" Value="0" />
+
+    <?if $(Include_NoAsserts) == True?>
+      <Variable Name="OptionsIncludeNoAsserts" bal:Overridable="yes" Persisted="yes" Value="1" />
+    <?else ?>
+      <Variable Name="OptionsIncludeNoAsserts" bal:Overridable="yes" Persisted="yes" Value="0" />
+    <?endif?>
```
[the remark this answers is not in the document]
mhegazy: changed

platforms/Windows/shared/shared.wxs
compnerd: It is easier to spot/read the paths that way. You are working with a path that looks like: `.../0.0.0+NoAsserts/usr/lib/swift/windows` which matches the `_noasserts_usr_lib_swift_windows`.

mhegazy: image

mhegazy:
closing in favor of https://github.com/swiftlang/swift-installer-scripts/pull/448

PR #426, opened 2025-05-21
Make installer window taller 

mhegazy:
To make room for additional options, this is a short term fix, to make the installer window taller. this is an alternative approach to #423.

The side bar art has solid colors, so extending it works without impacting image quality or aspect ratio.

here is what the result looks like:
image


mhegazy:
thanks @lxbndr ! appreciate it. yeah, the wix installer does not support svg natively so had to convert. i did commit both asserts though to make it easier for future changes

mhegazy:
screenshot in the description updated to match latest changes and new asset.

compnerd (approved):

etcwilde (approved):

PR #427, opened 2025-05-29
Rename bld.msi and bld.cab to bld.asserts.msi and bld.asserts.cab

mhegazy:
Restructure our toolchain projects to reflect which toolchain variant they are carrying. In this change we are renaming `bld.msi` and `bld.cab` to `bld.assers.msi` and `bld.asserts.cab`. Also moving the `.wixproj` to variant specific folder, to make room for adding additional variants in the future. 

Renaming the properties holding upgradcode and product name for consistency. 

This change should not have any functional impact.

### DIFF
--- a/platforms/Windows/bld/bld.wxi
+++ b/platforms/Windows/bld/bld.wxi
@@ -1,18 +1,19 @@
-﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Include xmlns="http://wixtoolset.org/schemas/v4/wxs">
 
   <?define ToolchainRoot = $(ImageRoot)\Toolchains\$(ProductVersion)+Asserts?>
 
   <Package
       Language="1033"
       Manufacturer="!(loc.ManufacturerName)"
-      Name="!(loc.Bld_ProductName)"
-      UpgradeCode="$(BldUpgradeCode)"
+      Name="!(loc.BldAsserts_ProductName)"
+      UpgradeCode="$(BldAssertsUpgradeCode)"
       Version="$(NonSemVerProductVersion)"
       Scope="$(PackageScope)">
 
-    <Media Id="1" Cabinet="bld.cab" EmbedCab="$(ArePackageCabsEmbedded)" />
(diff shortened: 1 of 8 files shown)


platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -3,6 +3,7 @@
 <WixLocalization xmlns="http://wixtoolset.org/schemas/v4/wxl" Culture="en-US">
   <String Id="DowngradeError" Value="A newer version of [ProductName] is already installed." />
   <String Id="Bld_ProductName" Value="Swift Build Tools" />
```
mhegazy, line 5: note for reviewers. `Bld_ProductName` is used for feature selection in the options page; where as  `BldAsserts_ProductName` is the product name in the msi. the user would not see the `BldAsserts_ProductName`, but adding it for future proofing when we have other variants.

platforms/Windows/bld/bld.wxi
```diff
@@ -1,18 +1,19 @@
...
       Manufacturer="!(loc.ManufacturerName)"
-      Name="!(loc.Bld_ProductName)"
-      UpgradeCode="$(BldUpgradeCode)"
+      Name="!(loc.BldAsserts_ProductName)"
+      UpgradeCode="$(BldAssertsUpgradeCode)"
       Version="$(NonSemVerProductVersion)"
       Scope="$(PackageScope)">
 
-    <Media Id="1" Cabinet="bld.cab" EmbedCab="$(ArePackageCabsEmbedded)" />
+    <Media Id="1" Cabinet="bld.asserts.cab" EmbedCab="$(ArePackageCabsEmbedded)" />
```
compnerd, line 14: Should we use some macros to make it easier to set this up for templating for `Asserts` and `NoAsserts`?

platforms/Windows/shared/swift.en-us.wxl
compnerd: Should we consider putting that in as a comment?

mhegazy: I can do that once we have a second variant. it feels a moot point now

mhegazy:
closing in favor of https://github.com/swiftlang/swift-installer-scripts/pull/428

PR #431, opened 2025-06-11
Add installation packaging script for FreeBSD

CodingCarpincho:
Swift should be distributed to FreeBSD users as a .pkg file which can be installed using pkg, the system package manager. This allows for a better user experience because it handles installing all of the required dependencies as well as placing the compiler in the user's PATH.

To create a FreeBSD package, we should have a script that automates the process of creating one. This commit adds such a script. It is designed to be invoked by the CI system, but can also be invoked by an end user.

### DIFF
--- a/platforms/FreeBSD/makePackage
+++ b/platforms/FreeBSD/makePackage
@@ -1,0 +1,141 @@
+#!/bin/sh -e
+#
+# Builds a FreeBSD package out of a toolchain.
+# Usage: makePackage /path/to/toolchain /path/to/final/package
+
+# Validate the command line arguments.
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 /path/to/toolchain /path/to/final/package" >/dev/stderr
+    exit 1
+fi
+
+# Create some variables for ease of use.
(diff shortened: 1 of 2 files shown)


shahmishal (approved):

shahmishal:
Thank you!

platforms/FreeBSD/makePackage
```diff
@@ -0,0 +1,115 @@
...
+echo "Generating manifest..."
+cat > "$METADATA_STAGING/manifest" <<EOF
+{
+  "name": "swift",
+  "version": "$SWIFT_VERSION",
+  "desc": "The Swift programming language",
+  "arch": "$(uname -m)",
+  "prefix": "/usr/local",
+  "origin": "",
+  "comment": "",
```
michael-yuji, line 92: nit: this should be a one-liner description that will show up in `pkg info`

CodingCarpincho: For the description field, we already have “the Swift programming language.” Any suggestions on what a short, succinct comment about the package could be?

shahmishal (approved):

PR #428, opened 2025-05-29
Rename toolchain msi and cab files to include name of variant (asserts)

mhegazy:
Follow up to https://github.com/swiftlang/swift-installer-scripts/pull/427. This mechanically applies the same changes to the rest of the toolchain msi's to have msi and cab name based on the variant they carry (i.e. asserts)

The `.wixproj` move to to variant specific folder, to make room for adding additional variants in the future. Also renaming the properties holding upgradcode and product name for consistency.

This change should not have any functional impact.

PR #432, opened 2025-06-26
Reflect variant name in directory ids

mhegazy:
This is a follow up change to #428. In this PR we are making directory IDs reflect the variant in the name of the install directory (e.g. `Swift\Toolchains\6.2.0+Asserts\usr`). 

There is no functional change in this change, it is just directory ids. Directory names and which components are installed to which directories should not be changed. 

This should make it easier to add new variants by following the same pattern, and will allow us to parametrize authoring that is shared between variants.

Naming convention:
- `ToolchainsVersioned` became `ToolchainsVersionedAsserts` -- this is our top-level toolchain install directory like `Swift\Toolchains\0.0.0+Asserts`
- Subdirectories like `_usr_share` became `toolchains_asserts_usr_share` to reflect the full path at install time i.e.``Swift\Toolchains\0.0.0+Asserts\usr\share`

### DIFF
--- a/platforms/Windows/bld/asserts/bld.asserts.wixproj
+++ b/platforms/Windows/bld/asserts/bld.asserts.wixproj
@@ -39,7 +39,7 @@
     <!-- FIXME(#81557) this needs to be properly staged once the sanitizers are included -->
     <HarvestDirectory Include="$(ImageRoot)\Toolchains\$(ProductVersion)+Asserts\usr\lib\swift\clang">
       <ComponentGroupName>SwiftStaticClangResources_asserts</ComponentGroupName>
-      <DirectoryRefId>_usr_lib_swift_static_clang</DirectoryRefId>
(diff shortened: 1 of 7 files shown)


mhegazy:
> Can you do a `s/toolchains_assert/toolchain_assert/`? The use is for a single toolchain installation right?

changed to use singular toolchain

compnerd:
Testing at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5949/

compnerd:
The installer built properly the test failures are unrelated.

mhegazy:
@compnerd this should be updated with changes in https://github.com/swiftlang/swift-installer-scripts/pull/419 now.

compnerd:
Testing this at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5977/

PR #433, opened 2025-06-26
Reflect variant name in component group ids

mhegazy:
This is a follow up change to #428. In this PR we are making component group IDs reflect the variant in the name they are carrying. 

There is no functional change in this change, changes are only to component ids.


compnerd:
The installer built successfully with this change; merging to unblock further work.

PR #440, opened 2025-07-22
WIX: mark the WinSDK components as Legacy

compnerd:
Begin marking the stable SDK as legacy. This is required to package up the new experimental runtime build in both dynamic and static formats.


compnerd:
CC: @mhegazy

mhegazy (approved):

etcwilde (approved):

PR #443, opened 2025-07-25
Update FreeBSD packaging script

CodingCarpincho:
We need to make two changes:

- First, the package `e2fsprogs-libuuid` is no longer shipped and the standard `libuuid` is used instead. We should therefore update the package manifest to reflect this change in the dependencies.
- Second, hard-coding package versions into the script is not very scaleable and will be a maintainence burden. Instead, we should ensure that the minimum package version required is whatever the build machine running this script has.

### DIFF
--- a/platforms/FreeBSD/makePackage
+++ b/platforms/FreeBSD/makePackage
@@ -78,6 +78,20 @@ cd "$CURRENT_DIRECTORY"
 SWIFT_VERSION="$("$SWIFT_TOOLCHAIN_HOME/bin/swift" --version | grep -Eo 'version [0-9.]+' | sed 's/version //')"
 echo "Swift version being packaged is $SWIFT_VERSION"
 
+# Get the currently installed version for all of the packages that we need.
+# This script assumes that it was run on the build machine and therefore
+# that the packages on this machine correspond to those Swift was built
+# against.
+get_package_version()
+{
+  echo "$(pkg info "$1" | grep Version | sed 's/Version.*: //')"
+}
+
+LIBUUID_PKG_VERSION=$(get_package_version libuuid)
+LIBXML2_PKG_VERSION=$(get_package_version libxml2)
+PYTHON_PKG_VERSION=$(get_package_version python311)
+SQLITE3_PKG_VERSION=$(get_package_version sqlite3)
+
 # Create the manifest file. All fields below are required. pkg_create(3)
 # will add additional fields as part of generating the package.
 echo "Generating manifest..."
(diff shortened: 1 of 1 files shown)


platforms/FreeBSD/makePackage
```diff
@@ -78,6 +78,23 @@ cd "$CURRENT_DIRECTORY"
...
+  echo "$(pkg info "$1" | grep Version | sed 's/Version.*: //')"
+}
+
+LIBUUID_PKG_VERSION=$(get_package_version libuuid)
+LIBFFI_PKG_VERSION=$(get_package_version libffi)
+LIBXML2_PKG_VERSION=$(get_package_version libxml2)
+MPDECIMAL_PKG_VERSION=$(get_package_version mpdecimal)
+PYTHON_PKG_VERSION=$(get_package_version python311)
+READLINE_PKG_VERSION=$(get_package_version readline)
+SQLITE3_PKG_VERSION=$(get_package_version sqlite3)
```
compnerd, line 96: Do we still need SQLite? I thought we used swift-toolchain-sqlite now and thus do not need the system vended SQLite.

platforms/FreeBSD/makePackage
```diff
@@ -78,6 +78,23 @@ cd "$CURRENT_DIRECTORY"
...
+# that the packages on this machine correspond to those Swift was built
+# against.
+get_package_version()
+{
+  echo "$(pkg info "$1" | grep Version | sed 's/Version.*: //')"
+}
+
+LIBUUID_PKG_VERSION=$(get_package_version libuuid)
+LIBFFI_PKG_VERSION=$(get_package_version libffi)
+LIBXML2_PKG_VERSION=$(get_package_version libxml2)
```
compnerd, line 92: This one I am weary about. Can we pin this to the exact version? I learnt that this library is not ABI stable and does not follow standard versioning practices. We are pinned to the exact version of libxml2 on Windows as there were ABI changes that break FoundationXML.

platforms/FreeBSD/makePackage
```diff
@@ -93,32 +110,32 @@ cat > "$METADATA_STAGING/manifest" <<EOF
...
       "origin": "math/mpdecimal",
     },
     "python311": {
-      "version": "3.11.11",
+      "version": "$PYTHON_PKG_VERSION",
       "origin": "lang/python311"
     },
     "readline": {
-      "version": "8.2.13_2",
+      "version": "$READLINE_PKG_VERSION",
```
michael-yuji, line 134: I think this is probably from Python too

platforms/FreeBSD/makePackage
jakepetroules: Well the idea for swift-toolchain-sqlite is that it's for systems without a native package management system to provide sqlite... meaning Windows and Android.

So far on Darwin we've been using the system SQLite and on Linux we've still been using the copy provided by the package manager... I expect it would make sense for FreeBSD to do the same as Linux here.

jakepetroules (approved):

PR #447, opened 2025-08-01
[Draft][windows] add Python 3.10.1 to the installer

charles-zablit:


platforms/Windows/python/python.wxs
```diff
@@ -0,0 +1,159 @@
...
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\pyexpat.pyd" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\python.cat" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\python.exe" />
```
mhegazy, line 43: do we really need the executable? i thought that all lldb needed is the dll and the built in modules?

platforms/Windows/bundle/installer.wixproj
```diff
@@ -19,6 +19,7 @@
     <ProjectReference Include="..\bld\asserts\bld.asserts.wixproj" BindName="bld.asserts" />
     <ProjectReference Include="..\cli\asserts\cli.asserts.wixproj" BindName="cli.asserts" />
     <ProjectReference Include="..\dbg\asserts\dbg.asserts.wixproj" BindName="dbg.asserts" />
+    <ProjectReference Include="..\python\python.wixproj" BindName="python" />
```
compnerd, line 22: Lets shuffle this down after all `ide` MSI. I wonder if it makes sense to rename this to match the 3-letter naming.

platforms/Windows/bundle/theme.xml
```diff
@@ -59,20 +59,21 @@
...
-            <Checkbox Name="OptionsInstallAndroidSDKARM64" X="210" Y="345" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_arm64)</Checkbox>
-            <Checkbox Name="OptionsInstallAndroidSDKAMD64" X="210" Y="363" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_amd64)</Checkbox>
-            <Checkbox Name="OptionsInstallAndroidSDKARM" X="210" Y="381" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_armv7)</Checkbox>
-            <Checkbox Name="OptionsInstallAndroidSDKX86" X="210" Y="399" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_x86)</Checkbox>
+            <Checkbox Name="OptionsInstallEmbeddedPython" X="192" Y="165" Width="-11" Height="17" TabStop="yes" FontId="3">#(loc.EmbeddedPython_ProductName)</Checkbox>
```
compnerd, line 62: I think that this should come at the very end as it is not really part of the toolchain, it is a third party dependency we are installing.

platforms/Windows/platforms/windows/windows.wxs
```diff
@@ -9,6 +9,7 @@
       Scope="$(PackageScope)">
 
     <?define PlatformRoot = "$(ImageRoot)\Platforms\Windows.platform"?>
+    <?define PythonRoot = "$(ImageRoot)\python"?>
```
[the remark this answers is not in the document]
compnerd: +1

platforms/Windows/python/python.wixproj
```diff
@@ -0,0 +1,5 @@
+<Project Sdk="WixToolset.Sdk/4.0.5">
+  <PropertyGroup>
+    <OutputName>python</OutputName>
```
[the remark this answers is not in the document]
compnerd: Do we really need 1 build per variant? I think it should be possible to share the Python installation across the different toolchain variants. We should lay them out such that they are isolated:

```
[INSTALLROOT]\Swift\Python-3.10.1\...
```

If we layout python this way, it would be isolated from any system installations and should be possible to share across different variants of the toolchain.

platforms/Windows/python/python.wxs
```diff
@@ -0,0 +1,159 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Include xmlns="http://wixtoolset.org/schemas/v4/wxs">
+
+  <Package
+      Language="1033"
+      Manufacturer="!(loc.ManufacturerName)"
+      Name="$(VariantProductName)"
+      UpgradeCode="$(VariantUpgradeCode)"
+      Version="$(NonSemVerProductVersion)"
+      Scope="$(PackageScope)">
```
compnerd, line 10: How do we version this component? Should we version it as per the Python version or the Swift toolchain version?

platforms/Windows/python/python.wxs
compnerd: Yes, we should include thee executable. Ultimately, the issue with all this is that LLDB is meant to be usable as a scriptable debugger, including for post-mortem scenarios. In such a case, you want to be able to execute a script, which is going to require the interpreter.

platforms/Windows/python/python.wxs
```diff
@@ -0,0 +1,159 @@
...
+
+    <Media Id="1" Cabinet="$(VariantCabinetName)" EmbedCab="$(ArePackageCabsEmbedded)" />
```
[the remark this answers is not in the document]
charles-zablit: Fixed, thanks! However, I think the file structure in `swift-installer-scripts\platforms\Windows\python` is not correct: the subdirectory should not be named `asserts`. Do you have a suggestion of what this should be named? Or should I just adapt a flat directory structure?

platforms/Windows/platforms/windows/windows.wxs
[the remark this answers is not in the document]
charles-zablit: Fixed, thanks!

platforms/Windows/bundle/installer.wixproj
charles-zablit: I moved it down below `ide`. Should it be at the very bottom?

Regarding the 3-letter renaming, I'm not opposed to it, however I think it would be good to differentiate between embedded Python and regular Python if we ever decide to bundle/chain the full Python msi with the toolchain installer.

If we are sure we are never going to chain the full Python msi, then we could name this `py3`?

platforms/Windows/bundle/theme.xml
charles-zablit: Fixed, thanks 👍

platforms/Windows/python/python.wixproj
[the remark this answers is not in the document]
compnerd: The installation information should be queryable and you can extract that from the registry. This is basically what `tcrun` does as well.

platforms/Windows/bundle/installer.wixproj
compnerd: I am certainly curious about _why_ the differentiation matters.

Ultimately, the difference between the embedded and full python is something that I think that we should somewhat blur. The actual difference is the lack of an installer (not a problem), a slightly smaller standard library (is that truly a concern?), and the missing `pip` which we can inject. So, difference-wise it is pretty small.

They also do not author MSIs, which means that the chaining would be more complicated than if they provided a MSM/MSI.

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,11 @@
...
+  <?define VariantName = asserts ?>
+  <?define VariantUpgradeCode = $(PythonUpgradeCode)?>
+  <?define VariantProductName = !(loc.EmbeddedPython_ProductName)?>
+  <?define VariantCabinetName = python.asserts.cab?>
+  <?define ToolchainVersionedVariantDirectory = ToolchainVersionedAsserts ?>
+  <?define VariantEnvironmentComponentGUID = 30629e0c-b376-47bc-bedf-fefb7d4ca61d?>
+  <?define PythonRoot = "$(ImageRoot)\Python"?>
+
+  <?include ../python.wxi ?>
+</Wix>
```
compnerd, line 11: Missing newline

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,11 @@
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+  <?define VariantName = asserts ?>
+  <?define VariantUpgradeCode = $(PythonUpgradeCode)?>
+  <?define VariantProductName = !(loc.EmbeddedPython_ProductName)?>
+  <?define VariantCabinetName = python.asserts.cab?>
```
compnerd, line 5: I agree with you, the cabinet name should not have `.asserts` in the name. We do not build the python interpreter with assertions.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+      Scope="$(PackageScope)">
+
+    <Media Id="1" Cabinet="$(VariantCabinetName)" EmbedCab="$(ArePackageCabsEmbedded)" />
+
+    <WixVariable Id="SideBySidePackageUpgradeCode" Value="$(VariantUpgradeCode)" />
+    <FeatureGroupRef Id="SideBySideUpgradeStrategy" />
+
+    <ComponentGroup Id="EmbeddedPython">
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\libcrypto-1_1-arm64.dll" />
```
compnerd, line 19: Is this file included on both ARM64 and AMD64?

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\libffi-7.dll" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\libssl-1_1-arm64.dll" />
```
compnerd, line 27: Likewise

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+        <File Source="$(PythonRoot)\unicodedata.pyd" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\vcruntime140.dll" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\vcruntime140_1.dll" />
+      </Component>
```
compnerd, line 84: This brings up the question - do we want to bundle an additional copy of the VCRuntime? This is an interesting question because the toolchain itself also does depend on the runtime. There is a copy in the system. And now in python. The system one however will get security updates, this will not. Since part of the argument for this upgrade is "security" - we shouldn't be bundling them IMO.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+      UpgradeCode="$(VariantUpgradeCode)"
+      Version="$(NonSemVerProductVersion)"
+      Scope="$(PackageScope)">
+
+    <Media Id="1" Cabinet="$(VariantCabinetName)" EmbedCab="$(ArePackageCabsEmbedded)" />
+
+    <WixVariable Id="SideBySidePackageUpgradeCode" Value="$(VariantUpgradeCode)" />
+    <FeatureGroupRef Id="SideBySideUpgradeStrategy" />
+
+    <ComponentGroup Id="EmbeddedPython">
```
compnerd, line 17: Can the directory attribute be moved to the `ComponentGroup` instead of the `Component`? That would simplify the authoring here.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\_uuid.pyd" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\_zoneinfo.pyd" />
+      </Component>
+    </ComponentGroup>
+
+    <Feature Id="EmbeddedPython" AllowAbsent="yes" Title="$(VariantProductName)">
```
compnerd, line 155: I don't think that we should allow this component to be absent. We would install nothing then.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,159 @@
...
+        <File Source="$(PythonRoot)\libffi-7.dll" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\libssl-1_1-arm64.dll" />
+      </Component>
+
+      <Component Directory="toolchain_$(VariantName)_usr_bin">
+        <File Source="$(PythonRoot)\LICENSE.txt" />
+      </Component>
```
compnerd, line 32: Why is the `LICENSE` file in `/usr/bin`? Shouldn't that be under `/usr/share/licenses`?

platforms/Windows/python/asserts/python.wxs
charles-zablit: Fixed, thanks 👍

platforms/Windows/python/python.wxi
charles-zablit: It's not, I've added the `ArchSuffix` define to address this. Thanks!

platforms/Windows/bundle/installer.wixproj
```diff
@@ -20,6 +20,7 @@
     <ProjectReference Include="..\cli\asserts\cli.asserts.wixproj" BindName="cli.asserts" />
     <ProjectReference Include="..\dbg\asserts\dbg.asserts.wixproj" BindName="dbg.asserts" />
     <ProjectReference Include="..\ide\asserts\ide.asserts.wixproj" BindName="ide.asserts" />
+    <ProjectReference Include="..\python\asserts\python.wixproj" BindName="python" />
```
mhegazy, line 23:
```suggestion
    <ProjectReference Include="..\python\asserts\python.asserts.wixproj" BindName="python.asserts" />
```

platforms/Windows/shared/swift.en-us.wxl
```diff
@@ -8,6 +8,7 @@
   <String Id="CliAsserts_ProductName" Value="Swift Command Line Tools (Asserts)" />
   <String Id="Dbg_ProductName" Value="Swift Debugging Tools" />
   <String Id="DbgAsserts_ProductName" Value="Swift Debugging Tools (Asserts)" />
+  <String Id="EmbeddedPython_ProductName" Value="Embedded Python 3.10.1" />
```
mhegazy, line 11: Since this is variant specific msi, we should have the name reflect that.

```suggestion
  <String Id="EmbeddedPythonAsserts_ProductName" Value="Embedded Python 3.10.1 (Asserts)" />
```

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,14 @@
...
+  <?define ToolchainVersionedVariantDirectory = ToolchainVersionedAsserts ?>
+  <?define VariantEnvironmentComponentGUID = 30629e0c-b376-47bc-bedf-fefb7d4ca61d?>
+  <?if $(ProductArchitecture) = "arm64" ?>
```
mhegazy, line 8: this is not variant specific. i would recommend moving it to the wxi to avoid it being duplicated for diffrent variants.

platforms/Windows/python/python.wxs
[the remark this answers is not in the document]
charles-zablit: I have tried to install the embeddable Python alongside the `Asserts` and `NoAsserts` directories, with `LLDB_PYTHON_HOME=../../Python` and tried other CMake defines like `LLDB_PYTHON_EXE_RELATIVE_PATH`.

This does not work because `lldb` still tries to look for `python310.dll` in the Path.

I see 2 possible solutions going forward:

# 1
Use Windows APIs to manually add specify the location of `python310.dll` before starting `lldb`. Since we optionally install Embeddable Python, this adds some hidden logic, if the user uses regular Python instead, in which case the DLL will actually be in the PATH. For this reason, I'm not in favor of this solution.


# 2
Put the contents of the embeddable Python folder in `usr/bin`.  We would have something like this:

```
Asserts/
├─ usr/
│  ├─ bin/
│  │  ├─ lldb.exe
│  │  ├─ python.exe
│  │  ├─ python310.dll
NoAsserts/
├─ usr/
│  ├─ bin/
│  │  ├─ lldb.exe
│  │  ├─ python.exe
│  │  ├─ python310.dll
```

I have tested this scenario and it works fine.

Now your concern [here](https://github.com/swiftlang/swift-installer-scripts/pull/447#discussion_r2254934627) was regarding this scenario:

```
Asserts/
├─ usr/
│  ├─ bin/
│  │  ├─ lldb.exe
│  │  ├─ python.exe
│  │  ├─ python310.dll
NoAsserts/
├─ usr/
│  ├─ bin/
│  │  ├─ lldb.exe
```

In this scenario, the `Asserts` variant works fine. The `NoAsserts` variant should work fine as well because `Asserts/usr/bin` will be in the PATH. If it's not, the user will have to install regular Python manually. I think this is OK because they would have had to install it either way. To I don't think that the following is true.

> so if we only install to one, the other variant is broken since lldb expects it to be in the path, and it will not be.


---

I think we should go with the 2nd solution, because it works with minimal changes to lldb's code.

platforms/Windows/bundle/installer.wixproj
compnerd: Marking this as resolved as I believe that we are currently planning to go with the shared installation.

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,14 @@
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+  <?define VariantName = asserts ?>
+  <?define VariantUpgradeCode = $(PythonUpgradeCode)?>
+  <?define VariantProductName = !(loc.EmbeddedPython_ProductName)?>
+  <?define VariantCabinetName = python.asserts.cab?>
```
compnerd, line 5:
```suggestion
  <?define VariantCabinetName = python.cab?>
```

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,14 @@
...
+  <?define ToolchainVersionedVariantDirectory = ToolchainVersionedAsserts ?>
```
compnerd, line 6: Do we need the `ToolchainVersionedVariantDirectory` define now?

platforms/Windows/python/asserts/python.wxs
```diff
@@ -0,0 +1,14 @@
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+  <?define VariantName = asserts ?>
```
compnerd, line 2: Do we need the `VariantName` now?

platforms/Windows/python/asserts/python.wxs
compnerd, line 5: I think that we should be able to drop the `Variant` in the variable names now.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,154 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Include xmlns="http://wixtoolset.org/schemas/v4/wxs">
+
```
compnerd, line 3: Extraneous whitespace

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,154 @@
...
+    <Media Id="1" Cabinet="$(VariantCabinetName)" EmbedCab="$(ArePackageCabsEmbedded)" />
+
+    <WixVariable Id="SideBySidePackageUpgradeCode" Value="$(VariantUpgradeCode)" />
+    <FeatureGroupRef Id="SideBySideUpgradeStrategy" />
+
+    <ComponentGroup Id="EmbeddedPython" Directory="toolchain_$(VariantName)_usr_bin">
+      <Component >
+        <File Source="$(PythonRoot)\libcrypto-1_1$(ArchSuffix).dll" />
+      </Component>
+
```
compnerd, line 21: I would drop the whitespace between the components.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,154 @@
...
+      Version="$(NonSemVerProductVersion)"
+      Scope="$(PackageScope)">
+
+    <Media Id="1" Cabinet="$(VariantCabinetName)" EmbedCab="$(ArePackageCabsEmbedded)" />
+
+    <WixVariable Id="SideBySidePackageUpgradeCode" Value="$(VariantUpgradeCode)" />
+    <FeatureGroupRef Id="SideBySideUpgradeStrategy" />
+
+    <ComponentGroup Id="EmbeddedPython" Directory="toolchain_$(VariantName)_usr_bin">
+      <Component >
```
compnerd, line 18:
```suggestion
      <Component>
```

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,154 @@
...
+        <File Source="$(PythonRoot)\_uuid.pyd" />
+      </Component>
+
+      <Component>
+        <File Source="$(PythonRoot)\_zoneinfo.pyd" />
+      </Component>
+    </ComponentGroup>
+
+    <ComponentGroup Id="EmbeddedPythonLicense">
+      <Component Directory="toolchain_$(VariantName)_usr_share_licenses">
```
compnerd, line 144: I don't think that we should be adding this to the toolchain layout but rather the python subdirectory.

platforms/Windows/shared/shared.wxs
```diff
@@ -55,6 +55,7 @@
                 <Directory Id="toolchain_asserts_usr_share_docc" Name="docc">
                   <Directory Id="toolchain_asserts_usr_share_docc_render" Name="render" />
                 </Directory>
+                <Directory Id="toolchain_asserts_usr_share_licenses" Name="licenses" />
```
compnerd, line 58: This should move into the directory hierarchy for Python.

platforms/Windows/readme.md
```diff
@@ -165,10 +167,11 @@ msbuild %SourceRoot%\swift-installer-scripts\platforms\Windows\bundle\installer.
   -p:Configuration=Release ^
   -p:BaseOutputPath=%PackageRoot%\online\ ^
   -p:ImageRoot=%ImageRoot%\Program Files\Swift ^
+  -p:PythonRoot=%ImageRoot%\Program Files\Python-3.10.1 ^
```
compnerd, line 170: I think that we want to sink this further 1 level as `%ImageRoot%\Program Files\Swift\Python-3.10.1`. When we do that, we can remove the parameter as we only need the python _version_ as we know the image root is `%ImageRoot%\Program Files\Swift` so we can compute the path relative to that as `$(ImageRoot)\Python-$(PythonVersion)`.

platforms/Windows/readme.md
```diff
@@ -165,10 +167,11 @@ msbuild %SourceRoot%\swift-installer-scripts\platforms\Windows\bundle\installer.
...
   -p:Platforms="android;windows" ^
   -p:AndroidArchitectures="aarch64;armv7;i686;x86_64" ^
   -p:WindowsArchitectures="aarch64;i686;x86_64" ^
-  -p:WindowsRuntimeARM64=%ImageRoot%\Prograam Files (Arm64)\Swift\Runtimes\0.0.0 ^
+  -p:WindowsRuntimeARM64=%ImageRoot%\Program Files (Arm64)\Swift\Runtimes\0.0.0 ^
```
compnerd, line 174: Hah! Thanks for fixing my typo.

platforms/Windows/python/python.wxs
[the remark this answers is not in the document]
mhegazy: I do not think option 2 is a good one. Installing the component per variant is not the best path as we discussed earlier.

Here is the thing, we install `usr/bin` in the `PATH`. so if we put `python.exe` in `usr/bin` users will find it if they just run `python`. this is not diffrent from putting it in `../../Python` and then adding this to the `PATH` so that lldb find it.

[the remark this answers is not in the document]
mhegazy: I think I have another option. lldb needs 2 things, `python310.dll` since it was linked against it, and cannot load without it, and the rest of the embeded python installation to execute script, and find imports etc.. 

For the first, we would add a new component under [`LLDB` ComponentGroup](https://github.com/swiftlang/swift-installer-scripts/blob/fa3419ec761650d04be8d11b6c4c47b796a4277b/platforms/Windows/dbg/dbg.wxi#L32) for `python310.dll` - this will be in the `PATH` since the whole folder is, but it is not an issue since this is the dll and not the `python.exe`.

For the second, we would install the new python.msi to the `..\..\Python` as we discussed before, and set `LLDB_PYTHON_HOME=../../Python` as part of the installation. 

@compnerd what do you think?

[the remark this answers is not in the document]
compnerd: That would be problematic as the `python.exe` (interpreter) is then going to fail to find the DLL. I think that the best option here is to do a `LoadLibraryW` for `python3.10.dll` and in the failing case use `SetDllDirectory` to inject the additional library search path. I just want to verify that it does not change the DLL lookup order if we do that (as in would block the previous search order from being honoured).

platforms/Windows/bundle/installer.wixproj
```diff
@@ -21,6 +21,7 @@
...
```
mhegazy, line 24: please get rid of `asserts` subdirectory, and just have everything directelly under `python`

platforms/Windows/python/python.wixproj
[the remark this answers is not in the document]
charles-zablit: This is fixed now, we share one common Python for all variants. The python installation lives in `[INSTALLROOT]\Python-3.10.1\python.exe`.

platforms/Windows/python/python.wxs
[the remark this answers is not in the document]
charles-zablit: We will be using `SetDllDirectory` once the following PR is merged:
- https://github.com/swiftlang/llvm-project/pull/11571/

platforms/Windows/bundle/installer.wixproj
charles-zablit: > and the missing pip which we can inject. So, difference-wise it is pretty small.

[Embedded Python does not support `pip`](https://docs.python.org/3/using/windows.html#windows-embeddable).

I think the scope of this patch is to fix the issues with users being confused by the missing `python310.dll`, where lldb crashes immediately (without an error message when executed from powershell):
- https://github.com/llvm/llvm-project/issues/43432
- https://github.com/llvm/llvm-project/issues/47381
- https://github.com/llvm/llvm-project/issues/53646
- https://github.com/llvm/llvm-project/issues/58095
- https://github.com/llvm/llvm-project/issues/59524
- https://github.com/llvm/llvm-project/issues/74073
- https://github.com/llvm/llvm-project/issues/85764

I might be misunderstanding, but I don't see the benefits of having a constrained version of Python with an unsupported `pip` installation that is not in the `PATH`. It seems reasonable, to me, to expect that a user who wants to do `lldb` Python scripting has a fully configured Python on their system.

platforms/Windows/python/python.wxs
charles-zablit: I think we should version it per the Python version, if we don't change the Python version between Swift versions, there is no need to reinstall Python.

platforms/Windows/python/asserts/python.wxs
charles-zablit: Removed it, thanks!

platforms/Windows/python/asserts/python.wxs
charles-zablit: Fixed, thanks!

platforms/Windows/python/python.wxi
charles-zablit: Fixed

platforms/Windows/python/asserts/python.wxs
charles-zablit: Removed it, it was only used for the license.

platforms/Windows/readme.md
charles-zablit: Fixed, thanks 👍

platforms/Windows/python/python.wxs
```diff
@@ -0,0 +1,8 @@
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+  <?define VariantUpgradeCode = $(PythonUpgradeCode)?>
+  <?define VariantProductName = !(loc.EmbeddedPython_ProductName)?>
+  <?define VariantCabinetName = python.cab?>
+  <?define VariantEnvironmentComponentGUID = 30629e0c-b376-47bc-bedf-fefb7d4ca61d?>
```
compnerd, line 5: I think that we can inline these values and I don't think that we need the `Variant` prefix as we do not have variants.

platforms/Windows/bundle/theme.xml
```diff
@@ -76,7 +76,7 @@
             <Checkbox Name="OptionsInstallToolchain" X="192" Y="417" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="0">#(loc.Toolchain_ProductName)</Checkbox>
             <Checkbox Name="OptionsInstallAssertsToolchain" X="210" Y="435" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsIncludeNoAsserts and OptionsInstallNoAssertsToolchain">#(loc.Asserts_Toolchain_ProductName)</Checkbox>
             <Checkbox Name="OptionsInstallNoAssertsToolchain" X="210" Y="453" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsIncludeNoAsserts and OptionsInstallAssertsToolchain">#(loc.NoAsserts_Toolchain_ProductName)</Checkbox>
-           
+            <Checkbox Name="OptionsInstallEmbeddedPython" X="192" Y="471" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallDBG">#(loc.EmbeddedPython_ProductName)</Checkbox>
```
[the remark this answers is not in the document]
mhegazy: correct. a few changes are included in that:
- move it up so that it is immediately under `OptionsInstallDBG`
- indent it so it looks like a sub feature
- set the enableCondition to track `OptionsInstallDBG`, so if the user did not choose `OptionsInstallDBG` in the first place, we do not install python. 

And if we do that we should also change the install condition in `installer.wxs` from `InstallCondition="OptionsInstallEmbeddedPython = 1"` to be `InstallCondition="OptionsInstallEmbeddedPython = 1 and OptionsInstallDBG = 1"` this way we only install it if both options are set

platforms/Windows/shared/shared.wxs
```diff
@@ -35,6 +35,14 @@
     </DirectoryRef>
   </Fragment>
 
+  <Fragment>
+    <DirectoryRef Id="INSTALLROOT">
+      <Directory Id="Python" Name="Python-$(PythonVersion)" >
```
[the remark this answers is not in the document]
charles-zablit: Fixed, thanks!

platforms/Windows/bundle/installer.wxs
```diff
@@ -134,6 +142,13 @@
         </MsiPackage>
       <?endif?>
 
+      <MsiPackage
+        SourceFile="!(bindpath.python)\python.msi"
+        InstallCondition="OptionsInstallEmbeddedPython = 1 and OptionsInstallDBG = 1"
```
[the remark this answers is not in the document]
charles-zablit: All the other `and` in this file are lowercase.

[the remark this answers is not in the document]
compnerd: There are others in the file which are uppercase. I know that I ran into some issues with the case in MSBuild. It is unfortunate that they are not consistently case-insensitive.

platforms/Windows/python/python.wxi
```diff
@@ -0,0 +1,40 @@
...
+    Version="$(PythonVersion)"
+    Scope="$(PackageScope)">
+
+    <Media Id="1" Cabinet="python.cab" EmbedCab="$(ArePackageCabsEmbedded)" />
+
+    <WixVariable Id="SideBySidePackageUpgradeCode" Value="$(PythonUpgradeCode)" />
+    <FeatureGroupRef Id="SideBySideUpgradeStrategy" />
+
+    <ComponentGroup Id="EmbeddedPython" Directory="Python">
+      <Files Include="$(PythonRoot)/**" />
```
compnerd: Please explicitly list the files. This is to ensure that we know _exactly_ what is being shipped for licensing purposes.

compnerd (changes requested):

platforms/Windows/bundle/installer.wxs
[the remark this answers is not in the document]
charles-zablit: Should I switch this one to uppercase? I re-checked, all the other `and` in this file are lowercased.
- WiXCodeSigning.targets
- bld.wxi
- installer.wixproj
- theme.xml
- windows.wxs

use uppercase.

Should we take this patch as an opportunity to sync `installer.wxs` to use uppercase as well?

compnerd (changes requested):

platforms/Windows/shared/shared.wxs
```diff
@@ -35,6 +35,389 @@
...
+                      <Directory
+                        Id="python_lib_site_packages_pip_25.3.dist_info_licenses_src_pip_v_u"
+                        Name="urllib3" />
+                    </Directory>
+                  </Directory>
+                </Directory>
+              </Directory>
+            </Directory>
+            <Directory Id="python_lib_site_packages_pkg_resources" Name="pkg_resources">
+              <Directory Id="python_lib_site_packages_pkg_resources_tests" Name="tests">
```
compnerd: Do we need to ship the test data and tests?

platforms/Windows/python/python.wxs
```diff
@@ -0,0 +1,3219 @@
...
+        <File Source="$(PythonRoot)\Scripts\wheel.exe" />
+      </Component>
+    </ComponentGroup>
+
+    <Feature Id="EmbeddedPython" AllowAbsent="no" Title="$(VariantProductName)">
+      <ComponentGroupRef Id="EmbeddedPython" />
+      <ComponentGroupRef Id="EmbeddedPythonPip" />
+    </Feature>
+  </Package>
+</Wix>
```
compnerd: Missing newline.

compnerd:
> Given that, I think we should reconsider the Wix v4 to v5 update (Wix 5 supports glob patterns).

I mean, if you want to take on the entire work of re-authoring the toolchain installer for WiX 5, that is an option. I don't want to adopt it and use just the one feature though. We should do a proper migration when ready.

> If one of those 1000 files changes (move, delete), the installer will not build and the toolchain build will fail.

That is kinda the idea. We don't want to ship something where the file might suddenly change and we are accidentally shipping something that we shouldn't be.

> We are relying on get-pip.py in build.ps1, which is not version controlled so we cannot pin to a specific version.

Ugh, if we cannot pin that, it is a problem.

I would be okay with doing this in two phases. We could get the initial bit in and add the pip inclusion in a follow up. That way we can at least partially make progress and hopefully get this in before the branch cut for 6.3.

compnerd (changes requested):

charles-zablit:
Tested and this build and works as expected.

PR #448, opened 2025-08-04
Add No-assert packaging support

mhegazy:
This changes introduces new toolchain variant for No-Assert. The authoring leverages our parametrization (https://github.com/swiftlang/swift-installer-scripts/pull/444) and only introduces a new set of project files and includes. 

**Note**: This does not change any of the existing Assert variant authoring or output (msi or cabs). The new msi's are only built if you pass `-IncludeNoAsserts`  to `build.ps1`. No changes on CI are needed at this point.

## Changes:
- New projects, and wxs with include files for bld, cli, dbg and ide. these result in producing new msi's `bld.noasserts.msi`, `cli.noasserts.msi`, `dbg.noasserts.msi` and `ide.noasserts.msi`
- Installer Options UI now has a new option to choose which variant to install, or both (see images below)
- Authoring for directories in shared.wxs to match the assert variant directories
- New upgrade codes for the new msi's
- New strings for the new msi's

## How does it work:
- If the bundler is built without `-IncludeNoAsserts`, noassert toolcahin is not included, and the user will see a grayed out entry that can not be changed. bundler will install asserts variant only, just like today (see first image below)
- If the bundler is built with `-IncludeNoAsserts`, both toolchain variants are included. the user has an option to install both (see second image), or only the noasserts variant (see third image). At least one has to be included. User can not be in a state where none is included (we ensure that through using enabled condition on the check boxes)
- If both variants are installed, the path will be set to the assert toolchain (to maintain backward compatibility). if only the noassert variant is chosen, the path will be set to the no-assert toolchain path.

## Screenshots

- Only Assert variant included:
Asserets-only

- Both variants included:
both

- Only No-Assert included:
noasserts-only


platforms/Windows/bundle/theme.xml
```diff
@@ -73,7 +73,9 @@
             <Checkbox Name="OptionsInstallAndroidSDKAMD64" X="210" Y="363" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_amd64)</Checkbox>
             <Checkbox Name="OptionsInstallAndroidSDKARM" X="210" Y="381" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_armv7)</Checkbox>
             <Checkbox Name="OptionsInstallAndroidSDKX86" X="210" Y="399" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_x86)</Checkbox>
-
+            <Checkbox Name="OptionsInstallAssertsToolchain" X="210" Y="417" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallNoAsserts and OptionsInstallNoAssertsToolchain">#(loc.Asserts_Toolchain_ProductName)</Checkbox>
```
compnerd: I don't understand this condition - why does installation of the asserts toolchain require the no asserts settings?

compnerd (commented):
Could you update the documentation as well please?

platforms/Windows/bundle/theme.xml
```diff
@@ -73,8 +73,9 @@
...
             <Checkbox Name="OptionsInstallAndroidSDKX86" X="210" Y="399" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_x86)</Checkbox>
-            <Checkbox Name="OptionsInstallAssertsToolchain" X="210" Y="417" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallNoAsserts and OptionsInstallNoAssertsToolchain">#(loc.Asserts_Toolchain_ProductName)</Checkbox>
-            <Checkbox Name="OptionsInstallNoAssertsToolchain" X="210" Y="435" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallNoAsserts and OptionsInstallAssertsToolchain">#(loc.NoAsserts_Toolchain_ProductName)</Checkbox>
+            <Checkbox Name="OptionsInstallToolchain" X="192" Y="417" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="0">#(loc.Toolchain_ProductName)</Checkbox>
+            <Checkbox Name="OptionsInstallAssertsToolchain" X="210" Y="435" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallNoAsserts and OptionsInstallNoAssertsToolchain">#(loc.Asserts_Toolchain_ProductName)</Checkbox>
```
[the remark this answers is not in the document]
mhegazy: That is the variable mapping to the build flag (`--includeNoAsserts`), i.e. is it even in the build or not. i think it should be "includeNoAsserts". will change

platforms/Windows/bld/noasserts/bld.noasserts.wxs
```diff
@@ -4,6 +4,7 @@
   <?define VariantUpgradeCode = $(BldNoAssertsUpgradeCode)?>
   <?define VariantProductName = !(loc.BldNoAsserts_ProductName)?>
   <?define VariantCabinetName = bld.noasserts.cab?>
+  <?define ToolchainVersionedVariantDirectory= ToolchainVersionedNoAsserts ?>
```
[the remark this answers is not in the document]
mhegazy: fixed

platforms/Windows/bundle/theme.xml
```diff
@@ -73,7 +73,10 @@
             <Checkbox Name="OptionsInstallAndroidSDKAMD64" X="210" Y="363" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_amd64)</Checkbox>
             <Checkbox Name="OptionsInstallAndroidSDKARM" X="210" Y="381" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_armv7)</Checkbox>
             <Checkbox Name="OptionsInstallAndroidSDKX86" X="210" Y="399" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsInstallAndroidPlatform">#(loc.Sdk_ProductName_Android_x86)</Checkbox>
-
+            <Checkbox Name="OptionsInstallToolchain" X="192" Y="417" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="0">#(loc.Toolchain_ProductName)</Checkbox>
```
compnerd, line 76: This feels wrong. The toolchain is all the bits (bld, cli, dbg, ide). Can we make this a label instead of a checkbox and indent the two variants below (as a poor-man's tree layout)?

platforms/Windows/bundle/theme.xml
```diff
@@ -73,7 +73,10 @@
...
+            <Checkbox Name="OptionsInstallAssertsToolchain" X="210" Y="435" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsIncludeNoAsserts and OptionsInstallNoAssertsToolchain">#(loc.Asserts_Toolchain_ProductName)</Checkbox>
```
compnerd, line 77: This condition still doesn't make sense to me - why does installing the asserts enabled toolchain require the no asserts toolchain?

platforms/Windows/bundle/theme.xml
```diff
@@ -73,7 +73,10 @@
...
+            <Checkbox Name="OptionsInstallNoAssertsToolchain" X="210" Y="453" Width="-11" Height="17" TabStop="yes" FontId="3" EnableCondition="OptionsIncludeNoAsserts and OptionsInstallAssertsToolchain">#(loc.NoAsserts_Toolchain_ProductName)</Checkbox>
```
compnerd, line 78: Why is this enabled by the installation of the asserts toolchain?

compnerd (commented):
Could you update the documentation as well please?

platforms/Windows/bundle/theme.xml
mhegazy: If the NoAsserts package is included in the msi (`OptionsIncludeNoAsserts`), and you check the NoAsserts variant checkbox (`OptionsInstallNoAssertsToolchain`) you are allowed to **uncheck** the Asserts variant (`OptionsInstallAssertsToolchain`).

The oposite is true as well. if the NoAsserts package are not included in the msi, or you did not check the NoAsserts, you can not uncheck the Asserts version. 

There has to be **one** toolchain variant selected.

platforms/Windows/bundle/theme.xml
mhegazy: see my comment above.

platforms/Windows/bundle/theme.xml
mhegazy: here is a video demonstrating how this should work:

https://github.com/user-attachments/assets/6095b45c-1e99-41b0-970b-3059167ee99e

compnerd: hm, should we default to the +Asserts and only allow selection/de-selection of the NoAsserts if included?

compnerd (approved):

PR #449, opened 2025-08-06
Simple correction

compnerd:
Remove unnecessary disk id specification and add the condition for the ARM64 slice of the experimental SDK.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6026/artifact/build/artifacts/

PR #450, opened 2025-08-07
WiX: attach disk IDs on directories when possible

compnerd:
Rather than adding the disk ID to each component we install, prefer to attach to the directory. This simplifies the authoring as these directories are known to be architecture agnostic and shared and we can avoid having to annotate each file.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6029/artifact/build/artifacts/

etcwilde (approved):

PR #453, opened 2025-08-17
WiX: Remove `PackageModelSyntax` dependency

xedin:
`PackageModelSyntax` module is about to get removed from Swift Package Manager because the functionality has been moved to swift-syntax. Sourcekit-LSP was the only external user of the library which is no longer the case.


xedin:
This is the last issue that prevents https://github.com/swiftlang/swift-package-manager/pull/9016 from landing.

compnerd:
Please link the cross repo build that built the toolchain - this is probably a good idea to smoke test.

PR #457, opened 2025-08-27
WiX: adjust the packaging rules for the SPM runtime

compnerd:
The package manager runtime installation adjusted for the split build installs the module with the target triple named thick module format. Adjust the packaging rules to integrate that into the image.


compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6076/artifact/build/artifacts/

mhegazy (approved):

PR #458, opened 2025-08-29
[WinSDK] Adjust the installer for platform modulemaps

egorzhdan:
This is a companion to https://github.com/swiftlang/swift/pull/83914.


compnerd (approved):
Please make sure to link the toolchain build to the PR and to smoke test the generated installer.

egorzhdan:
@compnerd I am attempting to test this installer, however, it doesn't seem to install swiftc.exe anywhere on my machine. `Windows.sdk` gets installed, but the entire `Toolchains` directory is missing. I tried running installer.exe with admin rights, also tried switching the installation directory to a custom one, that didn't help.

I don't think this is likely to be caused by my change, I suspect it's broken on the mainline as well.

PR #466, opened 2025-10-03
WiX: use the right variable name in feature install condition

mhegazy:
Changed in https://github.com/swiftlang/swift-installer-scripts/pull/461, we are using an undefined properties `Install...Platform` reverting to the correct property name `Install...SDK`


compnerd (approved):

compnerd:
Tested at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6139/artifact/build/artifacts/

PR #467, opened 2025-10-06
Use upper case for property names that are passed into the msi

mhegazy:
MSI uses property name case to define scope. Public properties (i.e., can be set from outside the MSI, e.g., command line) are all uppercase, and ones set within the installer are lowercase. See more in the [official docs](https://learn.microsoft.com/en-us/windows/win32/msi/public-properties?redirectedfrom=MSDN)

Since we are passing these properties to the MSI from the chainer, they need to be uppercase when they are being read. I have also changed the setting side to make sure we are consistent and do not run into this issue again.


compnerd (approved):
Please update the docs as well

mhegazy:
> Please update the docs as well

Do not think we expose any of the MSI properties in docs, do we?

mhegazy:
@compnerd did a round of local testing and it passed with latest changes. can you please trigger another build

PR #468, opened 2025-10-06
WiX: Fix MSI custom properties

Steelskin:
* MSI properties must be in CAPS_LOCK or they do not work properly.
* They must be introduced with a default value.
* Features should default to being installable (`Level=1`) and the override should set `Level` to 0 to skip installation of features. This is to prevent issues with uninstallation.

### DIFF
--- a/platforms/Windows/bundle/installer.wxs
+++ b/platforms/Windows/bundle/installer.wxs
@@ -188,12 +188,15 @@
           <MsiProperty Name="INSTALLROOT" Value="[InstallRoot]" />
 
           <MsiProperty Name="INSTALLARM64SDK" Value="[OptionsInstallWindowsSDKARM64]" />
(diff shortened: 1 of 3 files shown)


compnerd:
Installer built at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6148/artifact/build/artifacts/

compnerd (commented):
Case correction is the core of the change.

The handling of the levels is related to fixing the uninstall behaviour - the MSI must be invoked with the same property set on uninstall to take effect. We invert the values to get the proper behaviour. `Level` defaults to 1, and we set the value to `0` for the case where the SDK slice should be skipped.

The default values of the properties seem incorrect - we should default to these enabled.

compnerd (approved):

PR #470, opened 2025-10-08
WiX: remove support for packaging the legacy runtime

compnerd:
Clean up the legacy SDK packaging and remove support for it. Collapse all the various sites into the new layout only. Furthermore, rename the targets to remove the `Experimental` in the identifiers.


compnerd:
We no longer can package the legacy runtime as we do not build it.

PR #476, opened 2025-11-03
Windows: add property to verbose log when signing

speednoisemovement:
We're seeing flaky failures on CI from trusted signing. When I saw this locally, adding logging "fixed" it (suggesting a race condition). This change adds a property that lets us turn it on, both to see if it's sufficient and to learn something if it's not.

### DIFF
--- a/platforms/Windows/WiXCodeSigning.targets
+++ b/platforms/Windows/WiXCodeSigning.targets
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <SignOutput Condition=" '$(SignOutput)' == '' ">false</SignOutput>
     <SignOutput>$(SignOutput)</SignOutput>
+    <SignToolVerbose Condition=" '$(SignToolVerbose)' == '' ">false</SignToolVerbose>
   </PropertyGroup>
 
   <Target Name="FindSignTool">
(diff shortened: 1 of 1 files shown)


platforms/Windows/WiXCodeSigning.targets
```diff
@@ -44,8 +44,8 @@
       <SignToolPath Condition=" '$(SignToolPath)' == '' AND '$(PROCESSOR_ARCHITECTURE)' == 'ARM64' AND Exists('$(WindowsKitsRoot)bin\10.0.10240.0\arm64\signtool.exe')">$(WindowsKitsRoot)bin\10.0.10240.0\arm64\</SignToolPath>
 
       <!-- Microsoft recommends using their timestamp server for trusted signing: https://learn.microsoft.com/en-us/azure/trusted-signing/how-to-signing-integrations#:~:text=Trusted%20Signing%20certificates,microsoft.com/-->
-      <SignTool Condition=" '$(AzureSignMetadata)' != '' ">"$(SignToolPath)signtool.exe" sign /tr http://timestamp.acs.microsoft.com /fd sha256 /td sha256 /dlib "$(AzureSignDlib)" /dmdf "$(AzureSignMetadata)"</SignTool>
-      <SignTool Condition=" '$(AzureSignMetadata)' == '' ">"$(SignToolPath)signtool.exe" sign /tr http://timestamp.digicert.com /fd sha256 /td sha256 /f "$(CERTIFICATE)" /p "$(PASSPHRASE)" </SignTool>
+      <SignTool Condition=" '$(AzureSignMetadata)' != '' ">"$(SignToolPath)signtool.exe" sign /v /tr http://timestamp.acs.microsoft.com /fd sha256 /td sha256 /dlib "$(AzureSignDlib)" /dmdf "$(AzureSignMetadata)"</SignTool>
```
compnerd: While I get that this is important to be able to diagnose a current flake, I don't think that this is the right way. I would rather that we introduce a property to control this behaviour. Always printing the detailed logging makes it harder to understand the already difficult to process output.

speednoisemovement: Done

compnerd (approved):

compnerd:
Running a smoke test at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6220/. Once that passes we should be good to merge this.

compnerd:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6220/artifact/build/artifacts/

PR #478, opened 2025-11-10
WiX: ingest the MSI redistributable in the toolchain

compnerd:
With the ability to install the runtime with SxS manifesting, we now properly build the toolchain against a previous runtime which is isolated from the SDK associated runtime. In order to support this model, we must ingest the redistributable and package that into the toolchain.


compnerd:
We package a separate special SxS runtime now.

PR #508, opened 2026-02-20
WiX: package the swift-backtrace tool in the experimental SDK

compnerd:
The redistribution should include the backtrace tool. We have a hidden
knob to permit exclusion, but we do not surface this in the installer UI.
We expect most use cases to distribute the tool as part of the runtime
and would like to encourage its use.

### DIFF
--- a/platforms/Windows/rtl/shared/lib/rtl.shared.lib.wxs
+++ b/platforms/Windows/rtl/shared/lib/rtl.shared.lib.wxs
@@ -13,6 +13,7 @@
       <?define LibexecDirectoryComponentGuidGenerationSeed = "FCAC627D-44EC-4745-B81B-E12CE8F80951" ?>
       <?define RuntimeRoot = $(WindowsExperimentalRuntimeX86)?>
     <?endif?>
+    <?define SDKRoot = "$(ImageRoot)\Platforms\Windows.platform\Developer\SDKs\WindowsExperimental.sdk"?>
 
     <Directory Id="RUNTIMEDIR_$(ProductArchitecture)" ComponentGuidGenerationSeed="$(RuntimeDirectoryComponentGuidGenerationSeed)" />
     <Directory Id="LIBEXECDIR_$(ProductArchitecture)" ComponentGuidGenerationSeed="$(LibexecDirectoryComponentGuidGenerationSeed)" />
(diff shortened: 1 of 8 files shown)


mhegazy (approved):

compnerd:
CC: @al45tair

platforms/Windows/rtl/shared/lib/rtl.shared.lib.wxs
```diff
@@ -67,11 +67,9 @@
         <File Source="$(RuntimeRoot)\usr\bin\swift_Volatile.dll" />
       </Component>
 
-      <!-- TODO Enable when the experimental Runtimes build builds swift-backtrace
       <Component Directory="LIBEXECDIR_$(ProductArchitecture)">
-        <File Source="$(RuntimeRoot)\usr\libexec\swift\windows\swift-backtrace.exe" />
+        <File Source="$(RuntimeRoot)\usr\libexec\swift\swift-backtrace.exe" />
       </Component>
```
al45tair: Do we not need to disable this for 32-bit x86, as was done in `rtllib.wxs` in #509?

```suggestion
      <?if $(ProductArchitecture) != x86?>
        <Component Directory="LIBEXECDIR_$(ProductArchitecture)">
          <File Source="$(RuntimeRoot)\usr\libexec\swift\swift-backtrace.exe" />
        </Component>
      <?endif?>
```

compnerd: Ah, yes, we do.

compnerd: > Applying suggestions on deleted lines is currently not supported.

Well, that is annoying.

compnerd:
CI run at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6383/artifact/build/artifacts/

platforms/Windows/rtl/shared/lib/rtl.shared.lib.wxs
```diff
@@ -122,13 +117,26 @@
...
     </ComponentGroup>
 
+    <ComponentGroup Id="swift_backtrace_$(ProductArchitecture)" Directory="RUNTIMEDIR_$(ProductArchitecture)">
+      <!-- TODO(https://github.com/swiftlang/swift/issues/87499 ) enable this when we build this for x86 -->
```
etcwilde: I'm a little confused here. Above we have to wrap all of the `backtrace` containing component groups in `$(ProductArchitecture) != x86` so that they aren't empty, but here, we have several. How is that working? If it is supposed to work, can we do the same up on `rtlib.wxs` above?

compnerd: This definitely feels incorrect. I've adjusted the experimental side as well to mirror the legacy path. Thanks for catching this!

al45tair (approved):

compnerd:
Re-run at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6387/artifact/build/artifacts/

etcwilde (approved):

PR #518, opened 2026-04-07
WiX: package _PM-prefixed SwiftSyntax

speednoisemovement:
Needed after https://github.com/swiftlang/swift-package-manager/commit/35fd89aa7515499e9cb642ecbf1a6d96879ee527

Also removes outdated FIXME

### DIFF
--- a/platforms/Windows/cli/cli.wxi
+++ b/platforms/Windows/cli/cli.wxi
@@ -778,7 +778,30 @@
         <File Source="$(ToolchainRoot)\usr\bin\Workspace.dll" />
       </Component>
 
-      <!-- FIXME(compnerd) we should include the SPM import libraries -->
+      <Component>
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @DougGregor - it might be nice to support sharing the copy of swift-syntax across the compiler and SPM.

DougGregor:
Thank you for doing this. It is possible that I will end up reverting the change I made to SwiftPM, because I'm running into other issues with it further down the line.

speednoisemovement:
Thanks! Do you have an idea when you'll know for sure if you're reverting?

DougGregor:
I'm reverting the problematic change in the swift-package-manager: https://github.com/swiftlang/swift-package-manager/pull/9916 so we won't need this

PR #520, opened 2026-04-17
WiX: introduce file type associations for Swift

compnerd:
Thanks to the wonderful idea from @al45tair, including the icons, add some file type registrations for Swift files. This should associate a proper icon for the various files as well as provide a pretty file description in File Explorer.

The new `res` MSI (resources) homes the file type icons and explorer metadata. Introduce icons and file descriptions for:
- .swift
- .swiftmodule
- .swiftinterface
- .swiftdoc
- .swiftsourceinfo

The icons are installed to a version-independent path (`[INSTALLROOT]\Resources`) because the `DefaultIcon` registry values contain absolute paths to the `.ico` files. A versioned path would break icon resolution when a patch update removes the old directory.

Unlike the other MSIs, this package uses a standard MajorUpgrade rather than the SxS upgrade strategy. The content is identical across all Swift versions so there is no value in keeping multiple copies side-by-side. Every install of `res.msi` overwrites the previous one; the last uninstall cleans up.


compnerd:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6487/artifact/build/artifacts/

etcwilde (approved):
Icons are cute.

etcwilde:
image

Generated files are blue. Swift module, interface, and source info files have actual labels.

image

Source files have an orange Swift and say "Swift Source File" instead of "SWIFT file" under the "Type" field.

compnerd:
Going to merge this, I believe that @shahmishal was going to handle the MSI/CAB for the signing.

shahmishal:
We should wait to merge this, I wanted to get more clearly on icons. 

cc: @al45tair

compnerd:
Oh, oops, we can revert then - I didn't catch that.

PR #522, opened 2026-04-28
Add missing SwiftPM executables

owenv:
These should be included in the Windows distribution

Depends on https://github.com/swiftlang/swift-package-manager/pull/9999


jakepetroules (approved):

compnerd (approved):

owenv:
Adding a couple DLLs I missed

owenv:
Verified the executables launch correctly when using the toolchain from the latest run, merging

PR #527, opened 2026-05-26
Add _FoundationInternationalizationData library to Windows installer

jmschonfeld:
This duplicates the configuration of `_FoundationCShims` to add the `_FoundationInternationalizationData` library to the windows installer

### DIFF
--- a/platforms/Windows/platforms/FoundationInternationalizationDataHeaders.wxs
+++ b/platforms/Windows/platforms/FoundationInternationalizationDataHeaders.wxs
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+  <Fragment>
+    <ComponentGroup Id="_FoundationInternationalizationData" Directory="SDK_usr_include__FoundationInternationalizationData" Source="$(SDKRoot)\usr\include\_FoundationInternationalizationData">
+      <File Name="ListFormatData.h" />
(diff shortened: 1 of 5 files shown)


compnerd:
Do you expect the header set to grow much? If it is a single header, I feel that this is too much ceremony and we should just inline and duplicate the small (single currently) header set.

jmschonfeld:
I don't expect it to grow a _lot_, but over time I do expect it to grow comparably to the size of `_FoundationCShims`. Right now this library will only hold a single data table, but this is going to be the new home for CLDR data for Foundation so as internationalization APIs are written natively in Swift the number of tables / headers will likely grow.

I can remove the extra file if you prefer, but I did also find it nice that the list of headers is only in one place instead of needing to duplicate the header list in both places the file is referenced.

cc @itingliu @parkera

compnerd:
Right, that is precisely the reason for the extracted file. I think that it depends on the count of files we expect. If it is similar to the count of FoundationCShims, then this is likely the right approach.

jmschonfeld:
Holding on this for now until @compnerd can land a larger change to the installer scripts. Will merge the swift/swift-foundation PRs but wait to actually adopt the new library until we land this later

compnerd (approved):

compnerd:
Please ensure that we do a cross-repo toolchain test before merging

jmschonfeld:
Toolchain build passed successfully at https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6579/

PR #540, opened 2026-07-06
[WiX] Always sign catalogs with certificate

Steelskin:
Windows requires that the catalogs for the SxS assemblies be signed with the same certificate as the `publicKeyToken` referenced in the assembly manifests. This means that Azure Trusted Signing cannot be used for this case since Trusted Signing issues ephemeral short-lived certificates.

* Use a separate tool to sign the catalogs that always uses certificate + passphrase.
* Warn when no certificate is provided. This is because unsigned catalogs for SxS assemblies are rejected by Windows.


Steelskin:
Test run: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6647/

Steelskin:
I downgraded this to a warning because swift-CI does not set a CERTIFICATE.
New test run: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6652/

Steelskin:
Following offline discussions, we are dropping the Azure Trusted Signing flow as it cannot be used for signing native SxS catalogs.

PR #546, opened 2026-07-21
[6.4] WiX: include swift_RegexParser in the toolchain private runtime

compnerd:
This pull request updates the Windows installer configuration to include the new `swift_RegexParser` component in the toolchain. The main changes add the necessary directory and file references for this component.

**Installer configuration updates:**

* Added a new directory entry for `swift_RegexParser` under the toolchain's `usr/bin` section in `bld.wxi`.
* Registered the `swift_RegexParser.dll` file as a component to be installed, ensuring it is included in Windows builds.

PR #547, opened 2026-07-21
🍒 [6.4] WiX: Add reference to swift_RegexParser directory

Steelskin:
Follow-up from #535.

Cherry-pick of #538 (39a9cf777).


compnerd (changes requested):
Sorry, #546 already subsumes this.

PR #549, opened 2026-07-22
WiX: package FoundationNetworking as a SxS runtime

compnerd:
The Swift Package Manager uses Foundation Networking and thus the runtime must be bundled as a SxS private assembly. This should hopefully allow the toolchain to be usable.

Fixes: swiftlang/swift#90861


compnerd:
Run at: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows-arm64/80/artifact/build/artifacts/

Locally verified that this works as a bandaid. There are other failures and a more comprehensive fix is needed. But we should get this in to repair the 6.4 snapshots.

shahmishal (approved):

PR #550, opened 2026-07-22
WiX: introduce a new PRT MSI

compnerd:
This introduces a new PRT MSI which has two variants (`prt.asserts.msi` and `prt.noasserts.msi`) which provides the toolchain's private SxS runtime. There is an overlap in dependencies across different sets and IDE and CLI are independent but require the same runtimes. The new MSI allows us to install the proper set of libraries for the different installed components. It also simplifies the management of the private assemblies.


shahmishal (approved):

compnerd:
CI run: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6693/artifact/build/artifacts/

PR #554, opened 2026-07-28
WiX: Allow arbitrary signing command

speednoisemovement:
This introduces a new property `SignCommand` which allows overriding the full command used to code sign. If not provided, we fall back to the explicit `CERTIFICATE/PASSPHRASE` flow. 

Also removes Azure Artifact/Trusted Signing as a first-class option, but it can still be used via the `SignCommand` override.


platforms/Windows/RuntimeWin32Assemblies.targets
```diff
@@ -64,6 +64,6 @@
         Command="&quot;$(MakeCatToolPath)makecat.exe&quot; -v &quot;%(_RuntimeWin32Assembly.CatalogDefinition)&quot;" />
 
     <Exec Condition="'$(SignOutput)' == 'true'"
-          Command="$(SignTool) &quot;%(_RuntimeWin32Assembly.Catalog)&quot;" />
+          Command="$(SignCommand) &quot;%(_RuntimeWin32Assembly.Catalog)&quot;" />
```
compnerd, line 67: I don't understand this change

platforms/Windows/WiXCodeSigning.targets
```diff
@@ -54,19 +54,19 @@
...
-      <SignTool Condition=" '$(AzureSignMetadata)' != '' ">"$(SignToolPath)signtool.exe" sign $(VerboseFlag)/tr http://timestamp.acs.microsoft.com /fd sha256 /td sha256 /dlib "$(AzureSignDlib)" /dmdf "$(AzureSignMetadata)"</SignTool>
-      <SignTool Condition=" '$(AzureSignMetadata)' == '' ">"$(SignToolPath)signtool.exe" sign $(VerboseFlag)/tr http://timestamp.digicert.com /fd sha256 /td sha256 /f "$(CERTIFICATE)" /p "$(PASSPHRASE)" </SignTool>
     </PropertyGroup>
 
-    <Error Condition=" '$(SignToolPath)' == '' OR !Exists('$(SignToolPath)signtool.exe') "
+    <Error Condition=" '$(SignCommand)' == '' AND ('$(SignToolPath)' == '' OR !Exists('$(SignToolPath)signtool.exe')) "
            Text="Unable to locate signtool.exe. Set SignToolPath to the Windows SDK bin directory." />
+
+    <PropertyGroup>
+      <SignCommand Condition=" '$(SignCommand)' == '' ">"$(SignToolPath)signtool.exe" sign $(VerboseFlag)/tr http://timestamp.digicert.com /fd sha256 /td sha256 /f "$(CERTIFICATE)" /p "$(PASSPHRASE)"</SignCommand>
```
compnerd, line 68: Oh, this changes the time server - why?

platforms/Windows/RuntimeWin32Assemblies.targets
speednoisemovement: Renaming to be more accurate (it's the whole command, not just the tool)

platforms/Windows/WiXCodeSigning.targets
speednoisemovement: It's the previous non-trusted signing arm, same server:
```
      <SignTool Condition=" '$(AzureSignMetadata)' == '' ">"$(SignToolPath)signtool.exe" sign $(VerboseFlag)/tr http://timestamp.digicert.com /fd sha256 /td sha256 /f "$(CERTIFICATE)" /p "$(PASSPHRASE)" </SignTool>
```

platforms/Windows/RuntimeWin32Assemblies.targets
compnerd: Can we split that out into a separate change? I think that we should have three changes:
1. rip out the old azure signing
2. replace the command
3. add in the new hook point

speednoisemovement: > replace the command

Which? Do you just mean the rename? There's NFC for the cert/passphrase path here (beyond the ability to override it)

speednoisemovement: https://github.com/swiftlang/swift-installer-scripts/pull/555 for #1

speednoisemovement:
Superseded

PR #555, opened 2026-07-29
WiX: Remove trusted signing

speednoisemovement:
Azure Artifact/Trusted Signing is incompatible with the current toolchain build, since we need a stable publicKeyToken to embed in SxS manifests, and Artifact Signing certificates roll too often to support this.


compnerd (approved):
Thanks!

PR #558, opened 2026-08-03
WiX: Allow overriding SignCommand

speednoisemovement:
This change allows users to pass in a `SignCommand` property to override "vanilla" signtool.exe signing (for example, to use Azure Key Vault which requires using AzureSignTool instead of signtool.exe)

### DIFF
--- a/platforms/Windows/WiXCodeSigning.targets
+++ b/platforms/Windows/WiXCodeSigning.targets
@@ -54,17 +54,20 @@
   </Target>
 
   <Target Name="FindSignTool" DependsOnTargets="FindWindowsSDK">
-    <PropertyGroup>
+    <PropertyGroup Condition=" '$(SignCommand)' == '' ">
       <SignToolPath Condition=" '$(SignToolPath)' == '' ">$(WindowsSDKToolsPath)</SignToolPath>
       <SignToolPath Condition=" '$(SignToolPath)' != '' AND !HasTrailingSlash('$(SignToolPath)') ">$(SignToolPath)\</SignToolPath>
       <VerboseFlag Condition=" '$(SignToolVerbose)' == 'true' ">/v </VerboseFlag>
       <VerboseFlag Condition=" '$(SignToolVerbose)' != 'true' "></VerboseFlag>
-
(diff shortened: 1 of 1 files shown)


platforms/Windows/WiXCodeSigning.targets
```diff
@@ -54,17 +54,20 @@
...
       <SignToolPath Condition=" '$(SignToolPath)' == '' ">$(WindowsSDKToolsPath)</SignToolPath>
       <SignToolPath Condition=" '$(SignToolPath)' != '' AND !HasTrailingSlash('$(SignToolPath)') ">$(SignToolPath)\</SignToolPath>
       <VerboseFlag Condition=" '$(SignToolVerbose)' == 'true' ">/v </VerboseFlag>
       <VerboseFlag Condition=" '$(SignToolVerbose)' != 'true' "></VerboseFlag>
-
-      <SignCommand>"$(SignToolPath)signtool.exe" sign $(VerboseFlag)/tr http://timestamp.digicert.com /fd sha256 /td sha256 /f "$(CERTIFICATE)" /p "$(PASSPHRASE)" </SignCommand>
     </PropertyGroup>
 
-    <Error Condition=" '$(SignToolPath)' == '' OR !Exists('$(SignToolPath)signtool.exe') "
+    <Error Condition=" '$(SignCommand)' == '' AND ('$(SignToolPath)' == '' OR !Exists('$(SignToolPath)signtool.exe')) "
```
compnerd: Why this bit of change? `SignToolPath` being empty implies CWD no? This makes it harder - `-p:SignToolPath=".\"`.

speednoisemovement: It's just relaxing the pre-existing check to not require signtool if we passed in a custom command.

compnerd: But if you passed in `SignCommand`, ` '$(SignCommand)' == ''` would give `false` already. I guess I'm confused on the logic here.

speednoisemovement: With `SignCommand` passed in:
`$(SignCommand)' == ''` ->  evaluates false -> condition evaluates to `F ∧ (old condition)` -> F, so no error

Without `SignCommand` passed in,
`$(SignCommand)' == ''` ->  evaluates true -> condition evaluates to `T ∧ (old condition)` -> old condition

compnerd (approved):

PR #562, opened 2026-08-27
WiX: add new modulemap files to the manifest

compnerd:
The WinSDK module restructuring has introduced some new files which need to be enumerated for distribution.


etcwilde (approved):
Approved once https://github.com/swiftlang/swift/pull/91763 lands.

